PR #19, opened 2020-10-21
Adding the AgInspectionReport schema

brianwmunz:
This schema is defined for an agriculture inspection document.

### DIFF
--- a/docs/schemas/AgInspectionReport.json
+++ b/docs/schemas/AgInspectionReport.json
@@ -1,0 +1,72 @@
+{
+    "$id": "https://w3id.org/traceability/schemas/AgInspectionReport.json",
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "$comment": "{\"term\": \"AgInspectionReport\", \"@id\": \"https://w3id.org/traceability/AgInspectionReport\"}",
+    "title": "Agriculture Inspection Report",
+    "description": "Information on the Inspection and the observations made",
+    "type": "object",
+    "properties": {
+      "@context": {
+        "type": "array"
+      },
+      "type": {
+          "oneOf": [
+              { "type": "string" },
+              { "type": "array"}
+          ]
+      },
+      "facilityID": {
+        "$comment": "{\"term\": \"identifier\", \"@id\": \"https://schema.org/identifier\"}",
+        "title": "Facility Identifier",
+        "description": "Identifiers for an inspection facility.",
+        "type": "string"
+      },
+      "inspectorID": {
+        "$comment": "{\"term\": \"identifier\", \"@id\": \"https://schema.org/identifier\"}",
+        "title": "Inspector Identifier",
+        "description": "String of identifiers for inspectors.",
+        "type": "array"
+      },
+      "shipmentID": {
+        "$comment": "{\"term\": \"identifier\", \"@id\": \"https://schema.org/identifier\"}",
+        "title": "Shipment Identifier",
+        "description": "Identifiers for a shipment of goods.",
+        "type": "string"
+      },
+      "applicantID": {
+        "$comment": "{\"term\": \"identifier\", \"@id\": \"https://schema.org/identifier\"}",
+        "title": "Applicant Identifier",
+        "description": "Identifiers for the entity inspection applicant.",
+        "type": "string"
+      }, 
+      "inspectionLocation": {
+        "$comment": "{\"term\": \"Place\", \"@id\": \"https://schema.org/Place\"}",
+        "title": "Inspection Location",
+        "description": "Location of the inspection",
+        "$ref": "https://w3c-ccg.github.io/traceability-vocab/#place"
+      },
+      "inspectionDate": {
+        "$comment": "{\"term\": \"DateTime\", \"@id\": \"https://schema.org/DateTime\"}",
+        "title": "Inspection Date",
+        "description": "Date of the completed inspection.",
+        "type": "string"
+      },
+      "inspectionType": {
+        "$comment": "{\"term\": \"value\", \"@id\": \"https://www.gs1.org/voc/value\"}",
+        "title": "Inspection Type",
+        "description": "Type of Inspection.",
+        "type": "string"
+      },
+      "observation": {
+        "$comment": "{\"term\": \"observation\", \"@id\": \"https://schema.org/ItemList\", \"@type\": \"https://schema.org/ItemList\"}",
+        "title": "Observation List",
+        "description": "List of observations",
+        "type": "array",
+        "items": {
+          "$ref": "https://w3id.org/traceability/schemas/Observation.json"
+        }
+      }
+    },
+    "additionalProperties": false,
+    "examples": []
+  }


docs/schemas/AgInspectionReport.json
```diff
@@ -0,0 +1,72 @@
...
+        "description": "Identifiers for a shipment of goods.",
+        "type": "string"
+      },
+      "applicantID": {
+        "$comment": "{\"term\": \"identifier\", \"@id\": \"https://schema.org/identifier\"}",
+        "title": "Applicant Identifier",
+        "description": "Identifiers for the entity inspection applicant.",
+        "type": "string"
+      }, 
+      "inspectionLocation": {
```
OR13, line 42: wondering if this would be better as an inspection object, with a `location` property.

docs/schemas/AgInspectionReport.json
```diff
@@ -0,0 +1,72 @@
...
+      "@context": {
+        "type": "array"
+      },
+      "type": {
+          "oneOf": [
+              { "type": "string" },
+              { "type": "array"}
+          ]
+      },
+      "facilityID": {
```
OR13, line 18: wondering if this should also be a `facility` object of type `Organization` or similar, that might have an `id` property.

OR13: ping @msporny  @dlongley  any thoughts on this structure?

OR13 (changes requested):
I'd like to get some more feedback on the `concept` + `ID` naming convention.

We also might consider updating the associated spec generation code to include the updates, but that involves using npm / nodejs....

I think its reasonable to maybe attempt that after some dialog on the data model.

docs/schemas/AgInspectionReport.json
```diff
@@ -0,0 +1,72 @@
...
+        "$comment": "{\"term\": \"identifier\", \"@id\": \"https://schema.org/identifier\"}",
+        "title": "Facility Identifier",
+        "description": "Identifiers for an inspection facility.",
+        "type": "string"
+      },
+      "inspectorID": {
```
OR13, line 24: are these really strings or are the URLs? prefer URLs.

gjgd: The type says "array", I'm confused

brianwmunz: The array was a mistake sorry about that.  In terms of URLs...I'm not sure they would necessarily be URLs but maybe I'm not understanding.  I figured it would be an identifier that references the inspector.

docs/schemas/AgInspectionReport.json
brianwmunz: Given that an organization can have multiple facility locations we could have a location property in there?

docs/schemas/AgInspectionReport.json
mprorock: Agreed, I think a `location` property will be better here

docs/schemas/AgInspectionReport.json
mprorock: URLs pointing to how to retrieve an inspector is not a bad idea, but may not be feasible in all of our cases.  regularly we will just receive an alpha-numeric unique number that is assigned to that inspector

docs/schemas/AgInspectionReport.json
mprorock: I think an location property would be fine, but there will be a distinct id (alpha-num) assigned to that facility that we need to record as well

mprorock (changes requested):
I think if the changes around location properties are added, and the inspectorID issue is resolved, then this should be fine

OR13:
@mprorock wondering how you feel about the synthetic data generation code we have in this repo.

My intention was to provide a synthetic data generator for all RDF classes (types) in the vocab.

I have struggled with JSON-LD vocab before when the examples are "hand crafted" and never "validated / tested" with automation, but I can see javascript development skills being a burden to managing this at scale.

brianwmunz:
> I'd like to get some more feedback on the `concept` + `ID` naming convention.
> 
> We also might consider updating the associated spec generation code to include the updates, but that involves using npm / nodejs....
> 
> I think its reasonable to maybe attempt that after some dialog on the data model.

There's similar examples of this convention on schema.org.  For example: [productID](https://schema.org/productID) [propertyID](https://schema.org/propertyID) [taxID](https://schema.org/taxID) 
What do you think?

msporny:
> There's similar examples of this convention on schema.org. For example: productID propertyID taxID

Nitpicking... schema.org has a few rough edges and this is one of them. At the very least, we should follow industry standards for lower camel case for property names:

https://docs.scala-lang.org/style/naming-conventions.html

That would mean -- productId, propertyId, and taxId.

If we use non-URL identifiers, we should also be able to defend those choices. Yes, GS1 identifiers exist, but even those are going towards being resolvable and at a very minimum we can do URNs.

docs/schemas/AgInspectionReport.json
msporny: > wondering if this should also be a facility object of type Organization or similar, that might have an id property.

Agreed, especially if we're going to use something like a GLN to identify a facility (I hope so?). Ideally, we identify facilities using URLs so that we can add more linked data to that facility over time. Keeping these as non-URL identifiers won't help us enrich the data in the future.

msporny (changes requested):
Overall, very happy to see JSON Schemas for the JSON-LD data... this is going to really improve the quality of the data in the ecosystem.

The data modelling feels a bit off -- not by much, but in ways that make the data model a bit strange. Simple changes like `facilityID` -> `facility` would fix most of these data modelling issues.

The biggest concern I have is how much this deviates or aligns with the IBM Food Trust data model... have you done a comparison with that? Feels like we might be reinventing wheels that might have already been mostly invented elsewhere. From what I remember, their data model is a bit of a mess, but the same concepts would apply here. I wouldn't let that stand in the way of merging this PR... but do want to raise that as something that the industry would expect us to do from a completeness perspective.

brianwmunz:
Before I do the work, I figured I'd run this past everyone...
Based on what I'm seeing, I'm thinking what I should do is...

- Remove inspectorID, shipmentID, applicantID, inspectionDate, inspectionType, and facilityID from this schema
- Create an inspection property that references an inspection schema
- Create the inspection schema with the properties I removed.
- Change inspectorID to be an inspector property which references an inspector schema, and create that schema
- Change shipmentID to be a shipment property which references a shipment schema, and create that schema
- Same with applicantID and facilityID

Does this sound right?

OR13:
@brianwmunz I think thats right, it can feel a bit weird handling JSON-LD at first, so I will try and elaborate a bit more on why things look the way they do....

JSON-LD gets canonicalized as N-Quads, which are like triples... you can play around with possible expressions here:

- [here](https://json-ld.org/playground/)
- [here](https://www.jsonld-checker.com/?json=%257B%250A%2520%2520%2522%2540context%2522%253A%2520%255B%250A%2520%2520%2520%2520%2522https%253A%252F%252Fw3id.org%252Ftraceability%252Fv1%2522%250A%2520%2520%255D%252C%250A%2520%2520%2522type%2522%253A%2520%2522Observation%2522%252C%250A%2520%2520%2522property%2522%253A%2520%257B%250A%2520%2520%2520%2520%2522type%2522%253A%2520%2522ChemicalProperty%2522%252C%250A%2520%2520%2520%2520%2522name%2522%253A%2520%2522Titanium%2522%252C%250A%2520%2520%2520%2520%2522formula%2522%253A%2520%2522Ti%2522%252C%250A%2520%2520%2520%2520%2522inchi%2522%253A%2520%2522InChI%253D1S%252FTi%2522%252C%250A%2520%2520%2520%2520%2522inchikey%2522%253A%2520%2522RTAQQCXQSZGOHL-UHFFFAOYSA-N%2522%250A%2520%2520%257D%252C%250A%2520%2520%2522measurement%2522%253A%2520%257B%250A%2520%2520%2520%2520%2522value%2522%253A%2520%252236.424%2522%252C%250A%2520%2520%2520%2520%2522unitCode%2522%253A%2520%2522P1%2522%250A%2520%2520%257D%250A%257D)

the triples are of the form:

`<subject> <predicate> <object>`

The most common triples are `id` and `type`, so for example:

`inspection_id facility facility_id`
`inspection_id type AgInspection`

this is why you see so many arrays instead of objects in JSON-LD.... when you do see an object, which has an `id` its a node in an information graph, and the properties form predicates about that node.

brianwmunz:
Thanks for your patience on this (now and in the future) I'm relatively new to JSON-LD and I want to get it right. I'll take a shot at the changes I mentioned so we can review again. 
Thanks again!

mprorock:
> Before I do the work, I figured I'd float this...
> @OR13 @msporny @mprorock
> Based on what I'm seeing, I'm thinking what I should do is...
> 
> * Remove inspectorID, shipmentID, applicantID, inspectionDate, inspectionType, and facilityID from this schema
> * Create an inspection property that references an inspection schema
> * Create the inspection schema with the properties I removed.
> * Change inspectorID to be an inspector property which references an inspector schema, and create that schema
> * Change shipmentID to be a shipment property which references a shipment schema, and create that schema
> * Same with applicantID and facilityID
> 
> Does this sound right?

This sounds right to me.  

@msporny in relation to the other schemas that are out there in proprietary systems - yes, we are absolutely actively digging through all of existing specs.  One of the key issues that we are trying to address is modernizing much of this, as a lot of the specs are very outdated, or do not meet the needs of all the various parties we are balancing. We are actively in conversation with and reviewing items such as this ag inspection object with .gov parties, NGOs, commercial entities, as well as cross checking against the proprietary systems currently in the market (generally in domestic supply chain) to make sure we can interchange easily with those systems, while still allowing a clear path forward for standardization.  One of the issues we are seeing is that because this ecosystem (the ag track and trace side) is pretty fragmented, we tend to encounter a lot of proprietary definitions in existing APIs.  Where possible we are providing the ability to account for those extensions, without limiting core and required object properties that are mandated by all parties.  

The other issue we are running into, is that quite a few of the existing ag track and trace schemas are really just "normal" supply chain schemas or retail schemas that have been repurposed with ag branding on the public facing side, and don't actually account for the items that are required, especially for cross border transport.  Key in this, are the items related to the Ag inspection side of things which is one of the reasons we want to get the root objects that account for the peculiarities of ag inspections covered in a consistent and clean way that will allow mapping (or direct compatibility) to the parameters and data fields required by the various parties involved, as well as accounting for interchange with existing proprietary systems.

mkhraisha:
> @mprorock wondering how you feel about the synthetic data generation code we have in this repo.
> 
> My intention was to provide a synthetic data generator for all RDF classes (types) in the vocab.
> 
> I have struggled with JSON-LD vocab before when the examples are "hand crafted" and never "validated / tested" with automation, but I can see javascript development skills being a burden to managing this at scale.
> 
> However, if it makes the vocab better, I would be in favor of turning on CI checks for this, which would fail PRs like this in the future.... this would decrease the burden on editors, but increase the burden on other contributors.
> 
> cc @mkhraisha

I believe this is quite an important change to add, however I think this should probably be discussed as an issue rather than as a comment in an unrelated PR. I personally think having hand crafted VCs was a significant issue in our previous interop efforts and taking the steps to automate them was a very important move.

OR13:
@mkhraisha agreed, I opened https://github.com/w3c-ccg/traceability-vocab/issues/20

OR13:
The main action preventing this PR from being merged is the property name changes.

brianwmunz:
I think like I mentioned above in the list there's a bit of work I need to do to split it all out.  I'm just trying to get time to do so.

PR #40, opened 2020-11-18
added Oil and Gas specific use case descriptions

mkhraisha:
adding Oil and Gas specific use cases to the docs

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,6 @@
 {
     "editor.formatOnSave": true,
+    "prettier.enable": false,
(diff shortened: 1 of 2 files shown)


OR13 (changes requested):
This change set appears to have formatted the entire file.

You probably have prettier turned on in vscode :)

can you revert this to be just your additions? 

feel free to commit a `.vscode/settings.json` with prettier disabled to prevent this in the future.... I know this is frustrating I get bitten by this weekly :)

OR13 (approved):

mprorock (approved):

PR #51, opened 2020-12-08
Bringing schema refs local

### LINKED ISSUE #32: Redundancy with Schema.org
In creating new schemas for the AgInspectionReport, I noticed that in the current repo, we have schemas that are not modified or extended (as far as I can tell) from the ones they reference on GS1 or schema.org.  Why is that?  Shouldn't we simply ref out to Place, for example, rather than having it in the repo?

brianwmunz:
After discussion in issue #32 I've taken all $ref referenced to external schemas on schema.org and brought those schemas locally, including only the properties needed for the schemas in the traceability repo.  Also, this has required some general restructuring for existing schemas related to AgInspectionReport.json.
The new added schemas coming from schema.org are:
- Organization.json
- Person.json
- Product.json
- QuantitativeValue.json (for weight and/or size of Products)
- ParcelDelivery.json (for shipments of Products)

Also the Entity.json schema was created to describe an entity that could either be a Person or an Organization.

All testing was run and passed.

Fixes #32

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,10 @@ In order to have your contributions accepted you MUST:
 3. Run all tests locally and ensure they are all passing.
 4. Generate the latest version of the spec to include your changes to vocabular / data model.
 5. Open a Pull Request with your changes, a clear description of them in the description, and passing CI Tests.
+6. Any references to schemas you wish to include in your schema should be recreated locally and referenced there.  In these local schemas, only include the properties relevant to the schema you have created.  For example, [Person](https://schema.org/Person) is an existing schema on schema.org, but a Person.json schema has been added to this repo, including only the relevant and used properties for the other traceability schemas.  This is to hopefully make traceability schemas easier to understand and manage by not incorporating too many unnecessarily large schemas.  The schemas that have been made local in this way (like Person.json) should still reference the schema.org entry like so:
+```
+    "$comment": "{\"term\": \"Person\", \"@id\": \"https://schema.org/Person\"}",
+```
 
 #### Contributing to Vocabulary
 
@@ -71,6 +75,9 @@ and adding the path:
 5. review the latest spec changes by serving docs: `npx serve ./docs`.
 
 Follow the conventions established for the other properties, for example:
+
+###### Helpful Tips
+- Ensure 
 
 ###### Place
 
--- a/README.md
+++ b/README.md
@@ -76,9 +76,6 @@ and adding the path:
 
 Follow the conventions established for the other properties, for example:
 
-###### Helpful Tips
-- Ensure 
-
 ###### Place
 
 - [JSON Schema](./docs/schemas/Place.json)
--- a/docs/contexts/traceability-v1.jsonld
+++ b/docs/contexts/traceability-v1.jsonld
@@ -14,16 +14,16 @@
       "@id": "https://w3id.org/traceability/AgInspectionReport",
       "@context": {
         "facility": {
-          "@id": "https://schema.org/Place"
+          "@id": "https://w3id.org/traceability/schemas/Place.json"
         },
         "inspector": {
-          "@id": "https://w3id.org/traceability/schemas/inspector"
+          "@id": "https://w3id.org/traceability/schemas/Inspector"
         },
         "shipment": {
           "@id": "https://schema.org/ParcelDelivery"
         },
-        "applicantId": {
-          "@id": "https://schema.org/identifier"
+        "applicant": {
+          "@id": "https://w3id.org/traceability/schemas/Entity"
         },
         "inspectionDate": {
           "@id": "https://schema.org/DateTime"
@@ -70,6 +70,10 @@
         }
       }
     },
+    "Entity": {
+      "@id": "https://w3id.org/traceability/Entity",
+      "@context": {}
+    },
     "GeoCoordinates": {
       "@id": "https://schema.org/GeoCoordinates",
       "@context": {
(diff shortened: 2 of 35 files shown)


mprorock (approved):

mprorock:
thanks for extending the docs as well

mprorock:
@OR13 and @mkhraisha would like to merge this, any objections?

README.md
```diff
@@ -72,6 +76,9 @@ and adding the path:
 
 Follow the conventions established for the other properties, for example:
 
+###### Helpful Tips
+- Ensure
```
OR13, line 80: ?

docs/contexts/traceability-v1.jsonld
```diff
@@ -14,16 +14,16 @@
       "@id": "https://w3id.org/traceability/AgInspectionReport",
       "@context": {
         "facility": {
-          "@id": "https://schema.org/Place"
+          "@id": "https://w3id.org/traceability/schemas/Place.json"
```
OR13, line 17: this is not a valid definition.... see https://github.com/w3c-ccg/citizenship-vocab/blob/main/contexts/citizenship-v1.jsonld#L42

https://w3id.org/citizenship#commuterClassification

Note that `@id` should point to _human readable_ documentation for the term.

For example:

https://w3id.org/traceability/#facility-identifier

https://w3id.org/traceability/#place -> here was want this to be "https://schema.org/Place" since that is required to "use schema.org vocabulary".

docs/contexts/traceability-v1.jsonld
```diff
@@ -14,16 +14,16 @@
...
         },
         "inspector": {
-          "@id": "https://w3id.org/traceability/schemas/inspector"
+          "@id": "https://w3id.org/traceability/schemas/Inspector"
```
OR13, line 20: https://w3id.org/traceability/#inspector

docs/contexts/traceability-v1.jsonld
```diff
@@ -14,16 +14,16 @@
...
         },
         "shipment": {
           "@id": "https://schema.org/ParcelDelivery"
         },
-        "applicantId": {
-          "@id": "https://schema.org/identifier"
+        "applicant": {
+          "@id": "https://w3id.org/traceability/schemas/Entity"
```
OR13, line 26: https://w3id.org/traceability/#dfn-entities

docs/contexts/traceability-v1.jsonld
```diff
@@ -140,6 +144,66 @@
...
+        "products": {
+          "@id": "https://schema.org/ItemList"
+        }
+      }
+    },
+    "Person": {
+      "@id": "https://schema.org/Person",
+      "@context": {
+        "firstName": {
+          "@id": "https://schema.org/givenName"
```
OR13, line 188: notice how this term defintion leads to a human readable explanation

OR13 (changes requested):
A few issues with `@id` term URLs, but overall looks good.

please ensure the IRIs for `@id` always resolve to an html page with text describing the term.... if not page can be found, the term can be defined by this specification as a last resort...

In other words:

1. Search schema.org
2. Search GS1 Web Vocabulary

Search other ontologies...

The rule should be, never make JSON-LD worse.... I consider including a link to a term you have defined, that has no human readable defintion as making JSON-LD worse.

If some academic / other group defined the vocab IRI and you want to use it... thats work, but its IMO best to never support an IRI that does not define the term properly...

README.md
brianwmunz: oops.  removed.

brianwmunz:
thanks for the explanation @OR13 sorry about the confusion.  I went through and resolved the IDs to human readable content the best I could.  Thanks.
@mprorock  
@mkhraisha

docs/contexts/traceability-v1.jsonld
```diff
@@ -14,16 +14,16 @@
...
-          "@id": "https://w3id.org/traceability/schemas/inspector"
+          "@id": "https://w3id.org/traceability/#inspector"
         },
         "shipment": {
           "@id": "https://schema.org/ParcelDelivery"
         },
-        "applicantId": {
-          "@id": "https://schema.org/identifier"
+        "applicant": {
+          "@id": "https://w3id.org/traceability/#dfn-entities"
```
OR13, line 26: its possible to use anchor tags to remove the "dfn" part of this, for example:
https://github.com/w3c/did-spec-registries/blob/master/contexts/did-v1.jsonld#L21
https://www.w3.org/ns/activitystreams/#alsoKnownAs
https://www.w3.org/TR/did-core/#dfn-alsoknownas

https://www.w3.org/TR/did-core/#alsoKnownAs

OR13: this is non blocking, but we might want to address in README guidance

docs/contexts/traceability-v1.jsonld
OR13: left a comment in https://github.com/w3c-ccg/traceability-vocab/issues/53

docs/contexts/traceability-v1.jsonld
OR13: again a broken link, expect folks to open these

brianwmunz: i think this has been updated.

OR13 (approved):

brianwmunz:
Closing this PR after merging it into PR #52

PR #52, opened 2020-12-09
Ag inspection vc

### LINKED ISSUE #32: Redundancy with Schema.org

brianwmunz:
Updated the AgInspectionReport.test to create the VC credentials example.
While doing this I noticed that the Observation.json schema was throwing a context error due to the ID of measurement pointing to a non-existant schema.org schema.  Fixed this by pointing to the MeasuredValue.json local to the repo.  Also fixed a few other similar issues with contexts and naming.

After discussion in issue #32 I've taken all $ref referenced to external schemas on schema.org and brought those schemas locally, including only the properties needed for the schemas in the traceability repo. Also, this has required some general restructuring for existing schemas related to AgInspectionReport.json.
The new added schemas coming from schema.org are:

    Organization.json
    Person.json
    Product.json
    QuantitativeValue.json (for weight and/or size of Products)
    ParcelDelivery.json (for shipments of Products)

Also the Entity.json schema was created to describe an entity that could either be a Person or an Organization.

All testing was run and passed.

Fixes #32

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,10 @@ In order to have your contributions accepted you MUST:
 3. Run all tests locally and ensure they are all passing.
 4. Generate the latest version of the spec to include your changes to vocabular / data model.
 5. Open a Pull Request with your changes, a clear description of them in the description, and passing CI Tests.
+6. Any references to schemas you wish to include in your schema should be recreated locally and referenced there.  In these local schemas, only include the properties relevant to the schema you have created.  For example, [Person](https://schema.org/Person) is an existing schema on schema.org, but a Person.json schema has been added to this repo, including only the relevant and used properties for the other traceability schemas.  This is to hopefully make traceability schemas easier to understand and manage by not incorporating too many unnecessarily large schemas.  The schemas that have been made local in this way (like Person.json) should still reference the schema.org entry like so:
+```
+    "$comment": "{\"term\": \"Person\", \"@id\": \"https://schema.org/Person\"}",
+```
 
 #### Contributing to Vocabulary
 
@@ -71,6 +75,9 @@ and adding the path:
 5. review the latest spec changes by serving docs: `npx serve ./docs`.
 
 Follow the conventions established for the other properties, for example:
+
+###### Helpful Tips
+- Ensure 
 
 ###### Place
 
--- a/README.md
+++ b/README.md
@@ -76,9 +76,6 @@ and adding the path:
 
 Follow the conventions established for the other properties, for example:
 
-###### Helpful Tips
-- Ensure 
-
 ###### Place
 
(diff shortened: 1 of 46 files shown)


docs/test-vectors/Entity.json
```diff
@@ -0,0 +1,93 @@
...
+      "description": "Enhanced human-resource benchmark",
+      "address": {
+        "type": "PostalAddress",
+        "streetAddress": "2711 Nikki Keys",
+        "addressLocality": "South Ryannville",
+        "addressRegion": "Alaska",
+        "postalCode": "85791",
+        "addressCountry": "Poland"
+      },
+      "email": "Raymundo_Hodkiewicz@yahoo.com",
```
mprorock, line 18: let's use example.org and/or example.com for all email addresses

mprorock (changes requested):
let's remove apparent pi (event though it is not) such as example.org for all emails, and 555 numbers for all phone numbers

docs/test-vectors/Entity.json
brianwmunz: Okay, changes made

mprorock (approved):

OR13 (approved):
This a truely excellent PR

mprorock:
@mkhraisha any issues with merge?

OR13:
I have left comments on https://github.com/w3c-ccg/traceability-vocab/pull/51 seems like they may share some significant amount of code. I am happy to merge incrementally as long as we open issues for any "requested changes"

docs/test-vectors/Entity.json
OR13: see also https://github.com/w3c/did-core/pull/489#issuecomment-744544672

README.md
```diff
@@ -72,6 +76,9 @@ and adding the path:
 
 Follow the conventions established for the other properties, for example:
 
+###### Helpful Tips
+- Ensure
```
mprorock, line 80: Need this resolved or dropped before merge

mprorock (changes requested):

brianwmunz:
Mike and I decided it might be easiest to merge the changes from PR #51 into this one in order to ensure  there's no conflicts.  I've brought in the changes we made since submission of 51 so if we push this PR we should be all set with both and I can close 51.
Thanks.

OR13:
unclear on the order of merges for this, deferring #51 and #52 to @mprorock , merge when ready

mprorock (approved):

PR #55, opened 2020-12-15
Adding ppq report

brianwmunz:
Added schema for PPQ203 inspection form. The form itself was fairly straightforward, but I realized that we needed to start to accommodate subsets of test data generation. For PPQ for example, we don't want to pull in the full data for a Person schema, we just wanted their name.
To easily accomplish this, I've added json-mask (https://github.com/nemtsov/json-mask) to retrieve only the desired properties from the generated test content. It should be fairly straightforward.

NOTE
An earlier addition of a few schemas without the test vectors caused the test suite to fail. While I haven't fixed that issue here, I removed these files and ran the test against my files to ensure they were in compliance. Once the tests, passed, I added the files back in.


brianwmunz:
closing because I realized I made a mistake with the files from a previous PR

Issue #53, opened 2020-12-11
usage of IDs and Types in schemas


brianwmunz:
This question is more for clarification than anything else, but I wanted to make sure I was clear before getting too far...
In most of the existing schemas, the properties will be defined this way:
```
"identifier": {
      "$comment": "{\"term\": \"identifier\", \"@id\": \"https://schema.org/identifier\"}",
      "title": "Property Identifier",
      "description": "Identifiers for a property.",
      "type": "string"
},
```
The @id references the schema entry for GeoCoordinates and it's basically exactly what's defined in schema.org.  
As I'm building out larger schemas, there might be multiple identifiers (for example) in the same schema.  This will of course cause issues because you can't have two identical @id's in the same schema.  Also, given there are two identifier types in the same schema there is some sort of differentiation there that should be captured. 
This problem is especially relevant in having multiple places in the same schema (post of entry vs. foreign port export, for example)
All that to say, what I've done in an upcoming PR is to have the IDs reference the schema they're in and use @type to point to the schema type.  Like this (in a ppq203.json schema):
```
"certificateNumber": {
            "$comment": "{\"term\": \"certificateNumber\", \"@id\": \"https://w3id.org/traceability/schemas/ppq203.json#certificateNumber\", \"@type\": \"https://schema.org/identifier\"}",

...

},
        "carrierId": {
            "$comment": "{\"term\": \"carrierId\", \"@id\": \"https://w3id.org/traceability/schemas/ppq203.json#carrierId\", \"@type\": \"https://schema.org/identifier\"}",

...

},
```

Does this make sense as a proper way to define properties moving forward that are of a certain type but have a more specific definition in the schema?
@OR13 @mprorock @mkhraisha

OR13:
side note, you must space `@` in github to avoid pinging users who squat on obvious names like `@type` :)

This is a great question, and I don't have a great answer, but I will try and also ping @dlongley @msporny @tplooker 

```
@id
Used to uniquely identify things that are being described in the document with IRIs or blank node identifiers. This keyword is described in § 3.3 Node Identifiers.
...
@type
Used to set the data type of a node or typed value. This keyword is described in § 4.5 Typed Values.
```
id -> https://json-ld.org/spec/latest/json-ld/#node-identifiers
type -> https://json-ld.org/spec/latest/json-ld/#typed-values

`"@id": "https://w3id.org/traceability/schemas/ppq203.json#carrierId"` -> this looks wrong to me.

[...]

PR #60, opened 2020-12-21
AgProduct & AgPackage schemas

brianwmunz:
Added AgProduct and AgPackage schemas.  Generally what this PR is attempting to do is reflect the needed information on a typical [case label](https://www.producetraceability.org/documents/FINAL_PTI_Harmonized_Traceabilty_Case_Label_April_2020_(003).pdf) for inspection.  Given that one case could contain multiple products, I've created an AgProduct schema to represent the products within the case, and AgPackage to represent the case itself.

In both the AgProduct and AgPackage schemas and their dummy data generation, we have a property to hold a hash representing the binary of an image.  This will be used in VC.  In order to do this I've pulled in the Open Food Trust image, converted to binary and used crypto-js to hash it into sha256.


docs/contexts/traceability-v1.jsonld
```diff
@@ -14,10 +14,10 @@
       "@id": "https://w3id.org/traceability/AgInspectionReport",
       "@context": {
         "facility": {
-          "@id": "https://w3id.org/traceability/#facility"
+          "@id": "https://www.gs1.org/voc/Place"
         },
         "inspector": {
-          "@id": "https://w3id.org/traceability/#inspector"
+          "@id": "https://www.lawinsider.com/dictionary/food-inspector"
```
mprorock, line 20: I think we need to break food inspector out to its own type if an appropriate one does not exist in GS1, else change reference to GS1

docs/schemas/AgPackage.json
```diff
@@ -0,0 +1,76 @@
...
+            "description": "A four digit code digest of the GTIN code and Lot Code for use in integrating with voice picking systems.",
+            "type": "string"
+        },
+        "date": {
+            "$comment": "{\"term\": \"date\", \"@id\": \"https://schema.org/DateTime\"}",
+            "title": "Packing, Harvest, or Best By Date",
+            "description": "Human readable date describing the date the package was packed, harvested, or is 'best by.'",
+            "type": "string"
+        },
+        "labelImageUrl": {
```
mprorock, line 52: URL in the name is likely correct here, but let's make sure we don't want that to be URI in the future.  I am open on this, but we will be living with whatever this naming convention is for a while

mprorock (changes requested):
please double check usage of `@ id` - https://json-ld.org/spec/latest/json-ld/#node-identifiers by preference they will reference human readable page that has an underlying schema attached to it, but it is not clear to me from the spec if what is done here with links to non schema backed items is kosher or not.  @OR13 any thoughts as you have been working with json-ld longer than us?

mprorock:
to clarify a bit on the request, we may want to specify `@ type` explicitly, as per https://json-ld.org/spec/latest/json-ld/#example-28-example-demonstrating-the-context-sensitivity-for-type

docs/contexts/traceability-v1.jsonld
brianwmunz: I feel like [here](https://github.com/w3c-ccg/traceability-vocab/issues/53#issuecomment-743262231) @OR13 is saying to point to something human readable for explaining the property which seems to imply by referencing wikipedia that not having a schema backing it is okay.  BUT that being said, the link above is not my favorite either.  In fact, finding definitions of inspector was not easy.  So if we decided to break inspector out to its own type would we revert back to what I did before where we represent the definition that we're creating?    I'm good either way.

brianwmunz:
I feel like if we want to change the way the schemas generally use @type and @id we should make a separate PR because we'd have to change all of the existing schemas to the new format.  What do you think?

mprorock:
I see your point.  I am going to say let's run with it this way for now, but if we run into issues at least we know the alternative path.  Will change PR to approved

mprorock (approved):

Issue #69, opened 2021-02-02
Intent to Sell object


mprorock:
we need that ability to declare simply that there is a product that an organization intends to sell into some other destination country than the origin country.  This is not as complex as a product registration, but a very simple Org that will be exporting, target country to sell to, and the item that is intended for export.

OR13:
I suggest throwing JSON at this issue until something sticks, then finding and defining all necessary IRIs associated with such JSON.

bumblefudge:
The distinction between a product (GTIN-bearing, non-fungible) and commodity (fungible) might be the disconnect here.  E-Commerce only deals in the former, and I'm hearing from your use case the intent to sell the latter or, if the former, a high-level category of products rather than a pre-determined, finite set of products.  Our use-case is only looking at intent to sell 1 or more specific (and pre-registered) products.  Nobody throw JSON at me!

TallTed:
You might also find it helpful to explore [the GoodRelations vocabulary/ontology](http://wiki.goodrelations-vocabulary.org/Main_Page).

(Don't worry, @bumblefudge, it's not JSON. Unless you want it to be.)

mprorock:
@bumblefudge i think in this case we would ideally structure it that you could use a Product as referenced with as much detail as is available as early in the process.  In your case you would extend that out, as you have other legal reqs etc to cover, but yes, this will not necessarily be dealing with pre-registered products

mprorock:
resolved for now in #70

PR #70, opened 2021-02-03
Feature/intenttosell

mprorock:
Added IntentToSell base object to gover the most broad and generic use case where there is a seller, as possible buyer, and a product per #69


docs/schemas/IntentToSell.json
```diff
@@ -0,0 +1,55 @@
...
+                },
+                {
+                    "type": "array"
+                }
+            ]
+        },
+        "seller": {
+            "$comment": "{\"term\": \"seller\", \"@id\": \"https://w3c-ccg.github.io/traceability-vocab/#dfn-entities?v=1\"}",
+            "title": "Seller",
+            "description": "Entity that is declaring their intention for a product to be used at a differing location than where it was created.",
```
TallTed, line 25:
```suggestion
            "description": "Entity that is declaring their intention for a product to be used at a different location than where it was created.",
```

PR #71, opened 2021-02-03
Hotfix/legal entity identifier credential and le ievidence document

aaorlov:
Fixed schemas

LegalEntityIdentifierCredential:
- "leiCode" property changed as required

LEIevidenceDocument:
-  added "$comment" to property "data"

updated tests:
- changed date time format
- updated "faker.random.alphaNumeric" function usage

PR #72, opened 2021-02-04
merge to resolve conflicts and verify tests from hotfix on LEI

mprorock:
also resolves pr #71


mprorock:
merging as tests are passing and 43f83a5 and 010778e from pr #71 are accounted for

PR #82, opened 2021-02-09
(editorial) minor punctuation and whitespace edits

TallTed:

PR #92, opened 2021-02-10
punctuation and such; replaces #82

TallTed:


docs/index.html
```diff
@@ -4741,7 +4740,7 @@ <h2>Vocabulary</h2>
           "@id": "https://schema.org/unitCode"
         },
         "title": "Measurement Unit",
-        "description": "A string value indicating a Measurement Unit from UN/ECE Recommendation 20, Units of Measure used in International Trade e.g. GRM = gram - see <a href=\"http://www.unece.org/fileadmin/DAM/cefact/recommendations/rec20/rec20_rev3_Annex3e.pdf\">rec20_rev3</a>"
+        "description": "A string value indicating a Measurement Unit from UN/ECE Recommendation 20, Units of Measure used in International Trade, e.g., GRM = gram - see <a href=\"http://www.unece.org/fileadmin/DAM/cefact/recommendations/rec20/rec20_rev3_Annex3e.pdf\">rec20_rev3</a>"
```
mprorock, line 4743: any of these changes I believe will be overwritten by autogeneration.  the above core template item are fine.  @OR13  I know you commented on somehting similar recently

mprorock (commented):
if changes will be overwritten in this when generated from the schema files, they should be pulled from the commit.  verifying behaviour before approving

docs/index.html
TallTed: If they're going to be overwritten, then there's no need to pull them from the commit; they'll just be overwritten.

In such case, however, it _would_ be helpful to know where to target my fix, so that future overwrites bring correct punctuation instead of perpetuating the errors.

mprorock: :) sorry about that - Those are sourced from the schemas store in `docs/schemas/`

mprorock:
@OR13 @mkhraisha  any objection to me merging this?

OR13 (approved):

docs/index.html
OR13: yeah, I think the correct way to handle this would be for us to generate the html, and then run lint on it which would fix line length / indentation.

mkhraisha (approved):

docs/index.html
TallTed: @OR13 - Note that my changes here were not primarily about "line length / indentation" but punctuation, spelling, grammar, and the like. I've now applied my efforts to the `docs/schemas/` files, so the over-writes should now include the fixes I made here.

OR13: @TallTed thanks for these PRs, we really appreciate them!

PR #93, opened 2021-02-11
Phyto and schema updates

### LINKED ISSUE #83: dates should be in `YYYY-MM-DD` form
```suggestion
      "sellByDate": "10-3-2021"
```
I don't understand how changing dates falls under fixing IRIs ... but as long as you're fixing dates, please let's use ISO standard YYYY-MM-DD across the board!  (Is the above meant to say March 10, i.e., is it DD-MM-YYYY, or October 3, i.e., is it MM-DD-YYYY?)

_Originally posted by @TallTed in https://github.com/w3c-ccg/traceability-vocab/pull/81#discussion_r573217489_

Opening a new issue, per @OR13's request.

@mprorock cited the README:

> Wherever possible, dates should be formatted as `YYYY-MM-DD` so as to be directly compatible with `xsd:date`.

### LINKED ISSUE #84: vocabulary IRI should not link to PDFs
All IRIs should read to human readable documentation that is not a PDF (anchors or webpages, not pdfs).

### LINKED ISSUE #85: Fix definition for "Credential"
We are creating confusion by defining credential... we should instead define "TraceabilityCredential" or similar... VC Data Model is responsible for defining Verifiable Credentials... not us

brianwmunz:
There are five main things addressed with this PR:
- Added Phytosanitary schema with the appropriate tests
- Renamed the Credential schema (and all references to it) to Qualification.  This was done to avoid confusion between a credential relevant to an occupation and a verified credential.  
Fixes #85 
- Revisited the IDs of the Ag schemas to have them point to URIs backed by JSON and removed any references to PDFs  In the event that schema.org or GS1 was missing a definition, the property references the traceability schema.  
Fixes #84 
- We want to treat AgInspection as a subclass of Inspection report and ppq203 schema as a subclass of AgInspection.  This type of structure makes more sense to us than ref-ing to the super classes.  The issue right now is that "allOf" does not currently validate properly.  In the meantime until this is fixed, I moved all of the properties of AgInspection into ppq203 and Phytosanitary.  Along with this I reworked the test-vector generator scripts.
- Changed dates to YYYY-MM-DD formatting 
Fixes #83


mprorock (approved):

mprorock:
merging to correct a production issue on our side with some items referencing these schemas

PR #99, opened 2021-02-11
Add OpenAPI schema from JSON Schema example and auto-generation

mprorock:
This will generate an openapi schema file with basic get requests for each defined schema to make it easier for developers to cherry pick objects that they may need to use in an openapi definition or implementation


OR13 (approved):

mprorock:
@OR13 want me to merge this before we overhaul some test/generation naming, etc?

PR #79, opened 2021-02-09
Feature remaining vcs evidencedocs phase1 scope

aaorlov:

PR #102, opened 2021-02-12
Add cmtr

OR13:
Our data model has changed substantially since we first contributed `InspectionReport`... so I had to add a bunch of new stuff to make this valid.... I expect I will be in merge conflict hell shortly.

I also thing we should rename `test-vectors` to `generate-vocab`, and move the vc-js ci stuff into `generate-credential-examples`... I am happy to tackle that next.


mprorock:
> I also thing we should rename `test-vectors` to `generate-vocab`, and move the vc-js ci stuff into `generate-credential-examples`... I am happy to tackle that next.

This is a good idea.  We also should rename `help.js`

I have a big doc overhaul going as well. will note on the failing CI - have you looked into that? Looks like a few changes may need to be made @OR13

OR13:
@mprorock ci is fixed... this PR defines a lot of things I think other folks may want to use... in our next sync we should chat with the others about smaller types and `$ref` being your friend.

mprorock (approved):
gtg

mprorock:
@OR13 merge order preference between this and #79 - my inclination is to merge #79, then this one, but you may have to run one last merge on this pr before it can merge if I do that

OR13:
@mprorock I am fine fixing any conflicts, lets get other folks PRs in.

mprorock:
@OR13 cool, will merge once you commit your conflict fix

OR13:
These merge conflicts are getting pretty rough... we need to clean up the tests and build process.

mprorock:
agreed.  I reset after committing only the files i am intentionally changing and that helps a lot, but we need to clear this up from a documentation, test, and build process

PR #106, opened 2021-02-17
Update generate-spec.test.js

mprorock:
Fix a breaking issue on open api spec generation and validation

### DIFF
--- a/packages/traceability-vocab/src/generate-spec.test.js
+++ b/packages/traceability-vocab/src/generate-spec.test.js
@@ -115,7 +115,7 @@ it('should validate using json schema', async () => {
                   },
                 },
               },
-              500: 'Internal Server error',
+              500: { description: 'Internal Server error', }
             },
           },
         };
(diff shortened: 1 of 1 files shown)


packages/traceability-vocab/src/generate-spec.test.js
```diff
@@ -93,9 +93,21 @@ it('should validate using json schema', async () => {
       // only if everything validated with no errors should this add to the OpenAPI spec
       try {
         const $classComment = JSON.parse(classDefinition.schema.$comment);
+        
         openAPISpec.components.schemas[$classComment.term] = toOpenApi(
           classDefinition.schema,
         );
+        
+        const removeComments = function(obj) {
+          for(prop in obj) {
```
OR13, line 102:
```suggestion
          for(let prop in obj) {
```

OR13: or `// eslint-disable-next-line` everywhere...

mprorock: yeah, playing with a few options here on the recursion locally to get it to behave just right

OR13 (approved):

PR #112, opened 2021-02-17
Add bol take 2

OR13:
The PR adds Bill Of Lading and related data structures.... I expect this might not work for everyone yet, but it will let us start the conversation around how to handle this in a way that works for all vendors.


mprorock (approved):

mkhraisha:
Would like this to be held up till EOD, want to double check the BOL fields with business team

mkhraisha:
alternatively we can merge it and then open an issue for changes requested, idk which is preferable @OR13.

OR13:
@mkhraisha I would prefer to merge and then take PRs with changes... smaller change sets are easier to review... non of this should be considered final.

mkhraisha (approved):

PR #116, opened 2021-02-19
Dynamically register fixtures

OR13:
This will help avoid spelling errors, and makes it easier to create new entries.

This change only impacts tooling, not the spec.


mprorock (approved):
big thumbs up

PR #121, opened 2021-02-23
Additional AG VC examples

brianwmunz:
Added or updated VC tests for the following schema types:
- AgInspectionReport
- Entity
- ppq203
- Phytosanitary
- IntentToSell

PR #122, opened 2021-02-23
Adding a schema for Agricultural Activity Tracking

mprorock:


mprorock:
@OR13 @mkhraisha if we merge #121 first, i will merge those changes into this pr, then it should be good to go.  We would prefer a merge order on #121 then this one (#122) and then another small one coming on a few more inspections

mprorock:
passing ci, open for 24hrs, merging

PR #127, opened 2021-03-02
Refactor the build

OR13:
Now that CD is setup, we need to restructure files, so that PRs with changes can be smaller, and intermediate json can be ignored.

- [x] generate terminology as html from json at build time
- [x] separate schema validation and fixture generation 
- [x] separate open api generation from schema validation

This PR basically makes the spec generation a build time activity, and ignores all the intermediate data structures, so hopefully our PRs will get easier to review from here on out.

There are still warnings about missing test vectors, and now they show up in the spec as well.


OR13:
@mprorock @Therecanbeonlyone1969 this might be brutal to review, and I don't promise there won't need to be some after smaller fixes afterwords, but hopefully we will be able to rely on CI for building and CD for deploying, and no longer see the massive change sets of autogenerated files after this.

I expect there will need to be a few readme updates as well. since I moved everything around.

I would prefer to tackle that in another PR, since this one is already massive.

mprorock:
@OR13 I agree.  Think we need to get this in, make updates and issue smaller PRs after this one. I can bundle up all the readme items if you can create issues on them.

mprorock (approved):
Going to be downstream impact from this.  @OR13 I think an email describing the impacts to the supply chain cohort is in order given timing

mkhraisha (approved):

OR13:
I sent out an email, we are almost there fam.

OR13:
On second thought, lets do this one first, and I will implement @Therecanbeonlyone1969 's changes either in this PR, or immediately after.

mprorock:
Merging this since it will be critical to moving forward.  @OR13 has committed to implementing schema and non-ci related changes  provided by @Therecanbeonlyone1969  - this will negate pr #128 - but I will leave that open for now to reference - @OR13 @Therecanbeonlyone1969 please close #128 when appropriate

PR #133, opened 2021-03-06
Consolidate generation of VCs so that they are built directly from schemas

mprorock:


mprorock:
ok,I think this is now in a good place to move on from .  Still work to do on AJV and schema example validations, but I am going to open another PR on that so it doesn't get unwieldy.  The gist of this PR is that if you have a schema and example data generator, it is going to build all the right stuff for you.  It will also put an example VC from the "good" example in the spec doc for each top level term.

OR13 (approved):

OR13 (approved):

PR #135, opened 2021-03-07
Feature/doc improvements

mprorock:
prettied up some of the html generation, getting ready to overhaul some of the core docs for both developer and public facing - @OR13 @Therecanbeonlyone1969 @mkhraisha let me know if there are key areas you want in or out of docs

Things that are going to get in here:

- [x] format outbound html after generation
- [x] make sure readme in main schemas covers build process adequately
- [x] make sure top level readme covers goals, getting started, how to add a type, and build process
- [x] cover env variables related to build output useful for debugging
- [x] main spec to cover SMEs and Contacts for types of data
- [x] code best practices (e.g. not using dynamic dates, naming, inheritance, and expected level of denormalization)
- [x] some missing things that need help, e.g. allOf on the schema side
- [ ] basic flow diagram of vocab build process for a given element

PR #136, opened 2021-03-07
Feature/generator cleanup

mprorock:
@OR13 @Therecanbeonlyone1969 @mkhraisha I will be touching the following items today unless you say otherwise.  After this pass I will ensure that full generation through to issued VC example should be in for *most* of all listed schemas automatically, assuming that they have a generator.  I also will be addressing any items that have frequent data changes on generation that clutter up the build.  Please comment on this PR while in draft mode for anything you want to carve off and cover in your own PR.

Once we get to that point, with VC examples for everything, clean generation and validation, i am going to suggest we tag a version.

Generators that are too dynamic with data - these will get addressed first as they are currently mucking up fixtures with changes:

- [x]  LEIregistration
- [x] EcommerceWayBillRegistrationEvidenceDocument

Secondary offendors to this list:
- [x] LEIentity
- [x] LEIevidenceDocument
- [x] ShippingStop
- [x] ProductRegistrationEvidenceDocument
- [x] EcommerceBindingDataRegistrationCredential
- [x] EcommerceInvoiceRegistrationEvidenceDocument
- [x] EcommerceOrderRegistrationEvidenceDocument
- [x] EcommerceWayBillRegistrationCredential

Generators that mostly work but have now validation issues not currently handled (e.g. support for additional fields, allOf use, etc.):
- [x] AgActivity
- [x] AgInspectionReport
- [x] AgParcelDelivery
- [x] UsdaSc6
- [x] ppq203
- [x] ppq587

Generators that are absent or entirely broken:
- [ ] BillOfLading
- [ ] BillOfLadingCredential
- [x] CertifidMillTestReportCredential
- [ ] CommercialInvoiceCredential
- [ ] CrudeOilProduct
- [ ] QPInbond


Therecanbeonlyone1969:
@mprorock ... what does "too dynamic with data mean" from your PoV? I was under the impression we want to gen everything dynamically.

mprorock:
@Therecanbeonlyone1969 fixtures should be more or less fixed unless adding a new type, issue is that if dates for instance change when running the build, then fixtures that you did not actually touch as a part of your commit will be updated and the number of changed files will grow rapidly making it difficult if not impossible to review the actual changes that are being made.

mprorock:
@Therecanbeonlyone1969 basically we want the spec and all data to generate dynamically, but if there is no change to an object it should be repeatable.  the big offender on this is the way faker handles dates, that is the one item that random seeds don't really affect if you use past / present / future with faker

Therecanbeonlyone1969:
@mprorock ... we could hardcode dates in the generators and not use faker ... that should not be an issue. I think what you are saying is that CI/CD should only run on new or changed schemas/generators and not the rest ... I think that is an issue we should address but not by making things less dynamic. I am not deep enough into devops but I would say that .gitignore should be updated based on committed changes, no? so that is a module we can create. thoughts? cc @OR13

mprorock:
@Therecanbeonlyone1969 - no.  CI / CD runs on everything, and docs / spec / credentials are tested and re-generated with each run.  

basically, the `__fixtures__` folder should only see changes when:
- a new object is added (and then a folder for that object will appear, along with test fixtures)
- an object is actually changed
- something upstream from an object changed

I have already corrected the bulk of this.  There is now a master generators singleton in utils to provide consistent dates for use in generators, as well as a master instance of faker with a set seed.

mprorock:
going to convert this soon.  This addresses the date issues (as best i can tell) as well as fixes a lot of semi busted stuff.

I have stubbed or also made some adjustments in the following:
BillOfLading
CertifidMillTestReport
CommercialInvoice
CrudeOilProduct
QPInbond

A key note:  the schema is the master, so having a credential in the schemas doesn't really make sense, as the credential and then verified credential will get built automagically and programatically.  I am going to be going through and building out some saner generation on all the ag stuff to prevent fixutures from changing unless an upstream item changed, but will likely do this after plugfest.

jumping over to #135 now

OR13 (approved):

PR #139, opened 2021-03-09
chore: fix all http refs to schema.org - apologies, will touch a lot …

### LINKED ISSUE #138: replace all schema.org http with https IRI
https://github.com/w3c-ccg/traceability-vocab/blob/gh-pages/contexts/traceability-v1.jsonld#L5

possibly related to https://github.com/w3c-ccg/vc-http-api/issues/148

mprorock:
…of files

fixes #138


OR13:
oh boy... you just love pain don't you @mprorock  :)

mprorock:
also added an explicit check for the generated context against json-ld.  the implicit document loader checks are a generally ok, but will miss things that could cause issues with the context for other consumers

mprorock:
passing [ci](https://github.com/mesur-io/traceability-vocab/actions/runs/636339415), opening for review

OR13 (approved):

mprorock:
@OR13 not sure how much this will affect the other issue we are seeing in vc-http-api - however, i do not believe it can hurt.  do note - AJV is frustrating.  it does not like https for schema validation of JSON Schema, and the way we are handling schema -> ld adds some need for additional sanity checking in there, as well as modification of uri to http before running an sxplicit validate check on the schema obj itself against AJV

OR13:
merging this before it bites anyone.

PR #140, opened 2021-03-09
Typos in a couple of READMEs

nissimsan:


OR13 (approved):

packages/traceability-schemas/README.md
```diff
@@ -2,13 +2,13 @@
...
+Schemas should be named simply, and defined at a high a level as is reasonable.  Words such as 'credential' are redundant and should not be included in schema names.  e.g. a good schema name is `Product` for a type that represents data about a "product".  a BAD example would be `ProductCredential`.  A good sub-type that extends product with properties specific to some certain market would be `OrganicProduct`, assuming that an `OrganicProduct` had some specific attributes that are not represented in `Product`.  if the parent type accurately describes all attributes that are required, then a sub-type should not be created. If the only difference is a need to specify that the item is some specific "type" (in human english language meaning) of "Organic Product", then consideration should be given to adding a field to `Product` that specifies the "product type", such as `Product.productType`.  This prevents duplicate redundant objects and allows for sane querying and business rules handling off of common objects.
```
TallTed, line 11:
```suggestion
Schemas should be named simply, and defined at a high a level as is reasonable.  Words such as 'credential' are redundant and should not be included in schema names.  For instance, a good schema name for a type that represents data about a "product" is `Product`; a BAD example would be `ProductCredential`.  A good sub-type that extends `Product` with properties specific to some certain market would be `OrganicProduct`, assuming that an `OrganicProduct` had some specific attributes that are not represented in `Product`.  If the parent type accurately describes all attributes that are required, then a sub-type should not be created. If the only difference is a need to specify that the item is some specific "type" (in human english language meaning) of "Organic Product", then consideration should be given to adding a field to `Product` that specifies the "product type", such as `Product.productType`.  This prevents duplicate redundant objects and allows for sane querying and business rules handling of common objects. 
```

mprorock (changes requested):
showing some conflicts that need resolving.

nissimsan (commented):
Cool.

mprorock (approved):

Issue #64, opened 2021-01-26
Generalizing Use-Cases: Process VCs and non-process VCs


bumblefudge:
One helpful premise for aligning our use-cases for SC-specific test scenarios and user stories would be distinguishing process VCs (batch-specific, tied to specific import shipments) and non-process VCs (product definitions, legal entity verifications/enrolments, mappings of products to HTS codes, UPC codes, GTINs, etc). This has been very important in our use case, where there is a 1-to-many relationship between non-process VCs to shipments.

bumblefudge:
Reminder (as discussed on today's call): Update these use-cases with links if process/non-process distinction (or any other relevant change to the heirarchy or naming of credentials) and/or leave issue open if such changes are pending

OR13:
This was accomplished

OR13:
@TallTed can you review the related PR and make proposal for what is remaining to close this issue

OR13:
@bumblefudge can you propose next steps for closing this issue.

bumblefudge:
will do

TallTed:
This appears to me to be similar to internal vs external APIs, and along those lines I think there should be no built-in differentiation.

Any differentiation should be deployment-specific, i.e., business logic of the entity using the VCs.  That entity might apply different requirements to VCs issued and consumed by internal business processes, which might change over time due to acquisitions or spin-offs, where what was once "internal"/"process" becomes "external"/"non-process" and thenceforth has the latter requirements -- which might be satisfied if the original entity had treated both sets of VCs in the same way.

OR13:
pending a PR from @mkhraisha

TallTed:
Re-reading, this may be more about static vs variable VCs?  More detailed description will likely help with final resolution.

nissimsan:
This was referred to on the call as event-credentials. If that what this is(?), I'd recommend looking at this code list: 
https://service.unece.org/trade/uncefact/vocabulary/rec24/

mkhraisha:
I believe the vocabulary references is supposed to be EPCIS as outlined in #24

mprorock:
Per call: handle on a per-credential basis

Issue #24, opened 2020-10-29
What is the relationship between this vocabulary and EPCIS?


OR13:

mgh128:
The [GS1 Web vocabulary](https://www.gs1.org/voc) is an external extension to schema.org ( see http://blog.schema.org/2016/02/gs1-milestone-first-schemaorg-external.html ) and is primarily intended for expressing master data (e.g. ingredients, material composition, nutritional info, info about allergens, additives, certifications etc.) about products in greater detail than can currently be expressed using schema.org alone.

EPCIS is a GS1 open standard for interoperable exchange of visibility event data (for sharing traceability data), first released in 2007.  Core Business Vocabulary (CBV) is a complementary GS1 open standard that defines a number of code lists and values that should be used for populating some of the EPCIS event fields, such as bizStep and disposition, or expressing the type of source or destination or type of referenced business transaction.  Both of these GS1 open standards are available to download from https://www.gs1.org/epcis  where you can also access machine-readable artifacts such as XSD and WSDL files.  EPCIS and CBV are also recognised as ISO standards via the PAS process, as ISO/IEC 19987:2017 and  ISO/IEC 19988:2017 respectively.

Both EPCIS and CBV are being updated to version 2.0, which introduces a REST OpenAPI interface (to complement the existing WSDL+SOAP bindings) for the query and capture interface, as well as a JSON / JSON-LD data binding for the event data and instance/lot master data.  As part of that modernisation and the JSON-LD data binding, the CBV code lists and the EPCIS event data model will both be made available as browsable Linked Data models, probably with a very similar look-and-feel to the [GS1 Web vocabulary](https://www.gs1.org/voc) and with a JSON-LD download available for the underlying RDF data model, although we (GS1) feel that these are logically separate from the GS1 Web vocabulary, so they will have their own separate Linked Data URI namespaces hosted at gs1.org.

[...]

PR #163, opened 2021-05-12
to add note about process VCs in eComm use case

bumblefudge:
Down payment on #64 -- to be updated once process VCs multiply and again if/when process VCs get segregrated in a subfolder, prefixed, etc.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -449,6 +449,18 @@
               assessments, inferences, and data quality remediation on Customs
               pre-entry data.
             </p>
+            <p>
+              NOTE: To facilitate linking and alignment across internal and
+              external processes, many of the VCs modeled in this project are
+              labeled "process" VCs, meaning they refer to the internal
+              processes of a customs agency. Reusability here is likely minimal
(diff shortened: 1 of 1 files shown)


bumblefudge:
(happy to leave this open or close it and re-open later when there's more to do on this front.)

mprorock (approved):
I am OK with this for now - ultimately we will want to get this stuff resolved, then we can pull this note. It is however quite helpful for now.  if we get some tagging added to objects we can classify these types of items and then they will be less problematic

OR13:
@bumblefudge I think this is the best related issue to this: https://github.com/w3c-ccg/traceability-vocab/issues/24#issuecomment-840054844

if you want to add a respec issue link, you can do so like this:

```
   <p class="issue" data-number="24">
     Contributors are discussing how best to categorize process or event credentials vs product or organization credentials
    </p>
```

mkhraisha (approved):

OR13:
6 days open, multiple approvals, merging

PR #164, opened 2021-05-17
added simple root Event and fixed OGBillOfLading Generator

mkhraisha:
Adds a base "Event" Schema that captures very simple information. 

This schema should be extended by the following EPCIS event types:
* observation
* transformation
* aggregation
* disaggregation
* commission
* decommission

This PR is meant to spark conversation on what fields need to be in the base Event vs one of the Event Types

### DIFF
--- a/packages/traceability-schemas/schemas/Event.json
+++ b/packages/traceability-schemas/schemas/Event.json
@@ -55,7 +55,7 @@
         "products": {
             "$comment": "{\"term\": \"products\", \"@id\": \"https://schema.org/Product\"}",
             "title": "Products",
-            "description": "Products referenced by the event",
+            "description": "The products referenced by the event.",
             "type": "array",
             "items": {
                 "$ref": "https://w3id.org/traceability/schemas/Product.json"
--- a/packages/traceability-schemas/schemas/Event.json
+++ b/packages/traceability-schemas/schemas/Event.json
@@ -43,7 +43,7 @@
         "place": {
             "$comment": "{\"term\": \"place\", \"@id\": \"https://w3id.org/traceability#place\"}",
             "title": "Event Location",
-            "description": "Location where event took place",
+            "description": "The location where the event took place.",
             "$ref": "https://w3id.org/traceability/schemas/Place.json"
         },
         "eventTime": {
--- a/packages/traceability-schemas/schemas/Event.json
+++ b/packages/traceability-schemas/schemas/Event.json
@@ -49,7 +49,7 @@
         "eventTime": {
             "$comment": "{\"term\": \"eventTime\", \"@id\": \"https://schema.org/DateTime#v2\"}",
             "title": "Event Time",
-            "description": "Time when the event took place",
+            "description": "The time when the event took place.",
             "type": "string"
         },
         "products": {
--- a/packages/traceability-schemas/schemas/Event.json
+++ b/packages/traceability-schemas/schemas/Event.json
@@ -3,7 +3,7 @@
     "$schema": "https://json-schema.org/draft-07/schema#",
     "$comment": "{\"term\": \"Event\", \"@id\": \"https://schema.org/Event\"}",
     "title": "Organization",
-    "description": "An event such as transformation, aggregation, commission etc.",
+    "description": "An event such as a transformation, aggregation, commission, etc.",
     "type": "object",
     "properties": {
         "@context": {
--- a/packages/traceability-schemas/schemas/Event.json
+++ b/packages/traceability-schemas/schemas/Event.json
@@ -37,11 +37,11 @@
             "description": "The organization performing the activity.",
             "type": "array",
             "items": {
-                "$ref": "https://w3id.org/traceability/schemas/Organization.json"
+                "$ref": "https://w3id.org/traceability/schemas/Entity.json"
             }
         },
-        "eventLocation": {
-            "$comment": "{\"term\": \"eventLocation\", \"@id\": \"https://w3id.org/traceability#eventLocation\"}",
+        "place": {
+            "$comment": "{\"term\": \"place\", \"@id\": \"https://w3id.org/traceability#place\"}",
             "title": "Event Location",
             "description": "Location where event took place",
             "$ref": "https://w3id.org/traceability/schemas/Place.json"
@@ -53,7 +53,7 @@
             "type": "string"
         },
         "products": {
-            "$comment": "{\"term\": \"products\", \"@id\": \"https://www.gs1.org/voc/Product\"}",
+            "$comment": "{\"term\": \"products\", \"@id\": \"https://schema.org/Product\"}",
             "title": "Products",
             "description": "Products referenced by the event",
             "type": "array",
--- a/packages/traceability-schemas/schemas/Event.json
+++ b/packages/traceability-schemas/schemas/Event.json
@@ -28,7 +28,7 @@
         "eventId": {
             "$comment": "{\"term\": \"eventId\", \"@id\": \"https://schema.org/identifier\"}",
             "title": "Event Identifier",
-            "description": "Event Identifier.",
+            "description": "The Identifier of the Event.",
             "type": "string"
         },
         "actor": {
--- a/packages/traceability-schemas/schemas/Event.json
+++ b/packages/traceability-schemas/schemas/Event.json
@@ -22,7 +22,7 @@
         "eventType": {
             "$comment": "{\"term\": \"eventType\", \"@id\": \"https://schema.org/value\"}",
             "title": "Event Type",
-            "description": "Type of Event",
+            "description": "The Type of the Event.",
             "type": "string"
         },
         "eventId": {
--- a/packages/traceability-schemas/schemas/Event.json
+++ b/packages/traceability-schemas/schemas/Event.json
@@ -1,0 +1,67 @@
+{
+    "$id": "https://w3id.org/traceability/schemas/Event.json",
+    "$schema": "https://json-schema.org/draft-07/schema#",
+    "$comment": "{\"term\": \"Event\", \"@id\": \"https://schema.org/Event\"}",
+    "title": "Organization",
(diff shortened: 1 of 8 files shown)


packages/traceability-schemas/schemas/Event.json
```diff
@@ -0,0 +1,67 @@
...
+            "description": "Event Identifier.",
+            "type": "string"
+        },
+        "actor": {
+            "$comment": "{\"term\": \"actor\", \"@id\": \"https://w3id.org/traceability#Organization\", \"@type\": \"https://schema.org/ItemList\"}",
+            "title": "Actor",
+            "description": "The organization performing the activity.",
+            "type": "array",
+            "items": {
+                "$ref": "https://w3id.org/traceability/schemas/Organization.json"
```
mprorock, line 40: Do we want this to be an org? or is entity possibly better?  We at least also deal with individuals in the supply chain where Person sometimes is a better representation than Organization

packages/traceability-schemas/schemas/Event.json
```diff
@@ -0,0 +1,67 @@
...
+            "description": "Location where event took place",
+            "$ref": "https://w3id.org/traceability/schemas/Place.json"
+        },
+        "eventTime": {
+            "$comment": "{\"term\": \"eventTime\", \"@id\": \"https://schema.org/DateTime#v2\"}",
+            "title": "Event Time",
+            "description": "Time when the event took place",
+            "type": "string"
+        },
+        "products": {
```
mprorock, line 55: I am going to have to mull over use of Product here.  I think it is probably correct, but also thinking about raw materials becoming a product etc.

packages/traceability-schemas/schemas/Event.json
```diff
@@ -0,0 +1,67 @@
...
+                "$ref": "https://w3id.org/traceability/schemas/Organization.json"
+            }
+        },
+        "eventLocation": {
+            "$comment": "{\"term\": \"eventLocation\", \"@id\": \"https://w3id.org/traceability#eventLocation\"}",
+            "title": "Event Location",
+            "description": "Location where event took place",
+            "$ref": "https://w3id.org/traceability/schemas/Place.json"
+        },
+        "eventTime": {
```
mprorock, line 49: we may want to actually record a few more things here (possibly optionally), e.g. scheduled or planned start time, planned duration, actual start time, actual end time, etc.  open on this.

mprorock (changes requested):
Awesome start on this - I think the Event is going to be a a key interop point, and we will certianly need to issue VCs against certain events at least in our case on the ag side of things

OR13:
@mprorock  can you clearly state what changes need to be made in order to accept this PR?

potentially we can open issues to track them and iterate?

packages/traceability-schemas/schemas/Event.json
```diff
@@ -0,0 +1,67 @@
...
+            "$comment": "{\"term\": \"actor\", \"@id\": \"https://w3id.org/traceability#Organization\", \"@type\": \"https://schema.org/ItemList\"}",
+            "title": "Actor",
+            "description": "The organization performing the activity.",
+            "type": "array",
+            "items": {
+                "$ref": "https://w3id.org/traceability/schemas/Organization.json"
+            }
+        },
+        "eventLocation": {
+            "$comment": "{\"term\": \"eventLocation\", \"@id\": \"https://w3id.org/traceability#eventLocation\"}",
```
OR13, line 44: I am not sure we want to be defining "eventLocation"... can we use an existing ontology for this field?

packages/traceability-schemas/schemas/Event.json
```diff
@@ -0,0 +1,67 @@
...
+            "$ref": "https://w3id.org/traceability/schemas/Place.json"
+        },
+        "eventTime": {
+            "$comment": "{\"term\": \"eventTime\", \"@id\": \"https://schema.org/DateTime#v2\"}",
+            "title": "Event Time",
+            "description": "Time when the event took place",
+            "type": "string"
+        },
+        "products": {
+            "$comment": "{\"term\": \"products\", \"@id\": \"https://www.gs1.org/voc/Product\"}",
```
OR13, line 56: noting the use of a GS1 vocabulary item here, which is close to but not equal to schema.org product.

packages/traceability-schemas/src/generators/Event.js
```diff
@@ -0,0 +1,27 @@
+const faker = require('faker');
+const { getOrganization } = require('./Organization');
+const { getPlace } = require('./Place');
+const { getProduct } = require('./Product');
+
+const getEvent = () => {
+    const eventType = 'commission';
+    const eventId = '12345';
+    const actor = [getOrganization(), getOrganization()];
```
OR13, line 9: likely these helpers are injecting `@context`... please make sure we are not testing them in example data.

OR13 (changes requested):
https://github.com/w3c-ccg/traceability-vocab/pull/164/files#r634604678

^ this is the only critical objection I have, we need to be cleaning up our examples and avoid context nesting.

packages/traceability-schemas/schemas/Event.json
mprorock: +1 I think this should be a [Place](https://w3c-ccg.github.io/traceability-vocab/#Place)

packages/traceability-schemas/schemas/Event.json
mprorock: see note from @OR13 below - should this be GS1 or schema.org?  Other aspects of the vocab are utilizing schema.org

packages/traceability-schemas/schemas/Event.json
mprorock: I would prefer ability to record at a minimum

1. event start time
2. event end time

packages/traceability-schemas/schemas/Event.json
mkhraisha: I had misgivings on whether we should have two fields or just one for this, I wanted initially to have an `actor` and an `actingOrg` field. 

An entity might solve the problem. good shout

mprorock: I would prefer use of [Entity](https://w3c-ccg.github.io/traceability-vocab/schemas/Entity.json)

packages/traceability-schemas/schemas/Event.json
mprorock: > @mprorock can you clearly state what changes need to be made in order to accept this PR?
> 
> potentially we can open issues to track them and iterate?

noted above

packages/traceability-schemas/schemas/Event.json
mkhraisha: I think this makes sense to split up, probably another base object that encapsulates eventTime that would have the planned start/duration etc.

packages/traceability-schemas/schemas/Event.json
mkhraisha: absolutely should be place

packages/traceability-schemas/src/generators/Event.js
mkhraisha: theres two ways to fix this in my mind either:
 * we need a parameter to these helpers that indicates whether to inject context or not (my preference)
 * we need to strip context after its generated in each of these helpers

mprorock (approved):

mkhraisha:
Removed child context @mprorock  i deleted the generator and addressed most things in this conversation except for the event time stuff. would it be ok if i open an issue and assign it to you?

mprorock:
> event time stuff

let's not worry about the event time stuff as a blocker for now - we can add that as a later detail when required - probably as a subclass of this (e.g. event with duration or something like that)

OR13 (approved):

OR13:
On the call we discussed giving this one for day for reviews.

packages/traceability-schemas/schemas/Event.json
```diff
@@ -0,0 +1,67 @@
+{
+    "$id": "https://w3id.org/traceability/schemas/Event.json",
+    "$schema": "https://json-schema.org/draft-07/schema#",
+    "$comment": "{\"term\": \"Event\", \"@id\": \"https://schema.org/Event\"}",
+    "title": "Organization",
+    "description": "An event such as transformation, aggregation, commission etc.",
```
TallTed, line 6:
```suggestion
    "description": "An event such as transformation, aggregation, commission, etc.",
```

TallTed (changes requested):
Made many description strings more consistent with each other (start with article, close with period, etc.), particularly in ontology.  Some sample data is not good English (sentence fragments, primarily) but I'm not always sure what was intended, so some are now unchanged — but will want followup attention.

mkhraisha:
The remaining suggestions modify generated text, we probably need to actually dive into the generators and figure out how to fix them. This should probably be made its own issue and tracked accordingly. @OR13 @mprorock @TallTed

OR13:
These proposed changed look good to me, @mkhraisha can you approve them / reject them so we can see where any gaps are remaining?

packages/traceability-schemas/schemas/Event.json
TallTed: Consider "inputs" (raw materials, outputs from other processes/sources, etc.) and "outputs" (refined version of inputs becoming inputs to other processes/consumers, finished products, etc.).  These can then be easily turned into a complex *trace* from most-basic-origin (e.g., extraction of various ores from deep underground) to most-refined-consumable (e.g., a smartphone, tablet, or computer)...

OR13:
Feels like the generators may be getting in the way of having "realistic fixtures"... we should continue to discuss this on the related open issues.

TallTed:
Generated files should generally not be in git repos.  Whatever can be done to shift things so that the generated files go to github.io or similar would be best.

mkhraisha:
Apologies for the delays in tackling this. 
 Should be all good to go now.

OR13:
This looks good to me.

PR #171, opened 2021-06-28
Observation.date added attribute

nissimsan:


packages/traceability-schemas/schemas/Observation.json
```diff
@@ -30,6 +30,12 @@
       "title": "Measured Value",
       "description": "The measurement of an Observation.",
       "$ref": "https://w3id.org/traceability/schemas/MeasuredValue.json"
+    },
+    "date": {
+      "$comment": "{\"term\": \"date\", \"@id\": \"https://schema.org/Date\"}",
```
mkhraisha, line 35: This should probably be DateTime similar to existing schema calls, for example: 
[AgActivity](https://github.com/w3c-ccg/traceability-vocab/blob/fd05e51d64585908d2ef382a1304b449f757948c/packages/traceability-schemas/schemas/AgActivity.json#L44)
[IntentToSell](https://github.com/w3c-ccg/traceability-vocab/blob/fd05e51d64585908d2ef382a1304b449f757948c/packages/traceability-schemas/schemas/IntentToSell.json#L41)

mkhraisha (changes requested):

packages/traceability-schemas/schemas/Observation.json
mprorock: agreed

nissimsan: For sure, @mkhraisha. +1

mkhraisha (approved):

packages/traceability-schemas/schemas/Observation.json
```diff
@@ -30,6 +30,12 @@
       "title": "Measured Value",
       "description": "The measurement of an Observation.",
       "$ref": "https://w3id.org/traceability/schemas/MeasuredValue.json"
+    },
+    "date": {
+      "$comment": "{\"term\": \"date\", \"@id\": \"https://schema.org/DateTime\"}",
```
TallTed, line 35:
```
      "$comment": "{\"term\": \"date\", \"@id\": \"https://schema.org/DateTime\"}",
```
This is a *serious* misuse of `@id`, the value of which *should* be a unique-within-workspace identifier, *not* a datatype.  This PR may not be the best place to address this concern, but I thought it better to raise it now for redirection to the best place to address it, than to potentially forget about it.

OR13: @TallTed these schemas are used to [generate the JSON-LD context](https://github.com/w3c-ccg/traceability-vocab/blob/gh-pages/contexts/traceability-v1.jsonld), in this case, I would expect to see a generated context that looks like this (after some corrections):

```
 "date": {
      "@id": "https://schema.org/observationDate",
      "@type": "http://www.w3.org/2001/XMLSchema#dateTime" // or...  https://schema.org/DateTime ... depending on your ISO8601 vs W3C XSD Date Time religious views...
    },
```

See a related example from the W3C VC Data Model:

https://github.com/w3c/vc-data-model/blob/main/contexts/credentials/v1#L100

Changing the value of `@id` would break signatures, we try and NOT define terms differently than schema.org, or some more reputable vocabulary than ourselves:

https://schema.org/Observation

Does my suggestion look better?

TallTed: Right.  My misread was because I forgot how `"$comment"` is being abused by holding `@context` definitions without `@context` being mentioned anywhere nearby... So I was reading this as data, not as `@context`.

Too many idiosyncrasies across all the docs I'm reading. :-/

OR13 (approved):

OR13 (changes requested):
@nissimsan can you update to match my suggestion: https://github.com/w3c-ccg/traceability-vocab/pull/171/files#r662385324

OR13:
We discussed, this on the call, appears blocked the the fixtures related PR and the change request.

packages/traceability-schemas/schemas/Observation.json
nissimsan: @OR13, I understand this now and certainly agree. 
But this goes beyond observationDate, doesn't it? The whole Observation schema should point at https://schema.org/Observation (not a re-invention of ours). Making this change also and we can discuss if needed.

OR13 (approved):

OR13:
Please address merge conflicts

README.md
```diff
@@ -291,9 +291,11 @@ npm run test:schemas
    [schemas](./packages/traceability-schemas/schemas) folder.
 2. Add synthetic data generation for it to the
    [generators](./packages/traceability-schemas/src/generators).
-3. Run the build using: `npm run build:all`
-   Fix any errors found.
-4. Review the latest spec changes by serving docs: `npx serve ./docs`.
+3. OPTIONAL -- Add your new schema name to the file [includingevidence.json](./packages/traceability-schemas/src/data/vc/includingevidence.json) if you want to include the ```evidence``` property in your verifiable credential test fixture.
+4. OPTIONAL -- Add your new schema name to the file [includincredentialstatus.json](./packages/traceability-schemas/src/data/vc/includingcredentialstatus.json) if you want to include the ```credentialStatus``` property in your verifiable credential test fixture.
```
TallTed:
```suggestion
4. OPTIONAL -- If you want to include the ```credentialStatus``` property in your verifiable credential test fixture, add your new schema name to the file [includingcredentialstatus.json](./packages/traceability-schemas/src/data/vc/includingcredentialstatus.json).   
```

README.md
```diff
@@ -291,9 +291,11 @@ npm run test:schemas
    [schemas](./packages/traceability-schemas/schemas) folder.
 2. Add synthetic data generation for it to the
    [generators](./packages/traceability-schemas/src/generators).
-3. Run the build using: `npm run build:all`
-   Fix any errors found.
-4. Review the latest spec changes by serving docs: `npx serve ./docs`.
+3. OPTIONAL -- Add your new schema name to the file [includingevidence.json](./packages/traceability-schemas/src/data/vc/includingevidence.json) if you want to include the ```evidence``` property in your verifiable credential test fixture.
```
TallTed:
```suggestion
3. OPTIONAL -- If you want to include the ```evidence``` property in your verifiable credential test fixture, add your new schema name to the file [includingevidence.json](./packages/traceability-schemas/src/data/vc/includingevidence.json).
```

mprorock (approved):

OR13:
We agreed to merge this today.

PR #172, opened 2021-06-29
Initial ITI and Commodity

nissimsan:
First cut. Need to figure out context; also ITI examples aren't yet generated.

### DIFF
--- a/packages/traceability-schemas/schemas/CommodityType.json
+++ b/packages/traceability-schemas/schemas/CommodityType.json
@@ -1,0 +1,43 @@
+{
+  "$id": "https://w3id.org/traceability/schemas/CommodityType.json",
+  "$schema": "https://json-schema.org/draft-07/schema#",
+  "$comment": "{\"term\": \"CommodityType\", \"@id\": \"https://w3id.org/traceability#CommodityType\"}",
+  "title": "CommodityType",
+  "description": "A type of commodity, used to determine tarifs and required cross border documentation.",
+  "type": "object",
+  "properties": {
+    "@context": {
+      "type": "array"
+    },
+    "type": {
+      "oneOf": [
+        {
+          "type": "string"
+        },
+        {
+          "type": "array"
+        }
+      ]
+    },
+    "commodityDescription": {
+      "$comment": "{\"term\": \"commodityDescription\", \"@id\": \"https://w3id.org/traceability#commodityDescription\"}",
+      "title": "Commodity Description",
+      "description": "A description of the type of commodity.",
+      "type": "string"
+    },
+    "commodityCode": {
+      "$comment": "{\"term\": \"name\", \"@id\": \"https://w3id.org/traceability#commodityCode\"}",
+      "title": "Commodity Code",
+      "description": "The code representing the type of commodity",
+      "type": "string"
+    },
+    "commodityCodeScheme": {
+      "$comment": "{\"term\": \"description\", \"@id\": \"https://w3id.org/traceability#commodityCodeScheme\"}",
+      "title": "Description",
+      "description": "The codification scheme, such as WCO's 6-digit Harmonized System (HS) code listing or US Custom's Harmonized Tariff Schedule 10-digit (HTS) code listing.",
+      "type": "string"
+    }
+  },
+  "additionalProperties": false,
+  "examples": []
+}
--- a/packages/traceability-schemas/src/generators/CommodityType.js
+++ b/packages/traceability-schemas/src/generators/CommodityType.js
@@ -1,0 +1,14 @@
+const faker = require('faker');
+
+const getCommodityType = () => {
+  const example = {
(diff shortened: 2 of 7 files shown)


packages/traceability-schemas/schemas/CommodityType.json
```diff
@@ -0,0 +1,43 @@
...
+        {
+          "type": "string"
+        },
+        {
+          "type": "array"
+        }
+      ]
+    },
+    "commodityDescription": {
+      "$comment": "{\"term\": \"commodityDescription\", \"@id\": \"https://w3id.org/traceability#commodityDescription\"}",
```
OR13, line 23: it is best to avoid creating new terms, so if we can change this `@id` to point to schema.org, gs1 or another more reputable vocabulary we should.

I also like to avoid camel casing when possible, so I would prefer:

```
 "description": {
      "$comment": "{\"term\": \"description\", \"@id\": \"https://schema.org/description\"}",
```

packages/traceability-schemas/schemas/CommodityType.json
```diff
@@ -0,0 +1,43 @@
...
+      "title": "Commodity Description",
+      "description": "A description of the type of commodity.",
+      "type": "string"
+    },
+    "commodityCode": {
```
OR13, line 28: similar comment as above... 

maybe we can compact this to something like:

```
"code": {
   "value": "123",
   "scheme": "HTS"
}
```

packages/traceability-schemas/src/generators/CommodityType.js
```diff
@@ -0,0 +1,14 @@
+const faker = require('faker');
+
+const getCommodityType = () => {
```
OR13, line 3: This looks great, but I wonder if we ought to start using static JSON fixtures?

nissimsan: Oh right, not even using it. Copy-paste oversight.

packages/traceability-schemas/schemas/CommodityType.json
TallTed: This "jam a bunch of structured data into a nominally free-text value, where you must escape many characters along the way" thing is a recipe for disaster.  If it's going to persist a while (as it currently appears), the humans working with these specs will find it easier if some syntactically meaningless whitespace can be used to softly format that free-text, at least in any sample or test data presented in or with these specs, as in ---
```
 "description": {
      "$comment": "{ \"term\":  \"description\", 
                     \"@id\":   \"https://schema.org/description\"
                   }",
```

OR13: yeah, I agree... the issue is how to extend JSON Schema properly, instead of using the `hack` we invented.

See https://json-schema.org/specification.html

particularly:

- http://json-schema.org/understanding-json-schema/reference/generic.html

We've discussed changing the way this works, do you have any recommendations based on these links?

TallTed: Well, based on [this segment](http://json-schema.org/understanding-json-schema/reference/generic.html#id3) (emphasis below, mine), using `$comment` in this way is effectively forbidden!

>The `$comment` keyword is strictly intended for adding comments to a schema. Its value must always be a string. Unlike the annotations `title`, `description`, and `examples`, **_JSON schema implementations aren’t allowed to attach any meaning or behavior to it whatsoever, and may even strip them at any time._** Therefore, they are useful for leaving notes to future editors of a JSON schema, but should not be used to communicate to users of the schema.

I don't have a solution.  All I can really say is that this is a hazard of constant wheel reinvention (JSON Schema, here) by each new player in the software stack (JSON, here, with which I have many disagreements, not least being the lack of syntax for inline comments which generally should, if not must, be preserved, which really *can* enable self- and/or internally-documented code).

OR13: from a "JSON Schema" perspective, stripping JSON-LD is fine... the resulting document is still valid JSON Schema without these annotations.

We should explore adding a new property, and setting JSON Schema strict mode to false, that might be more readable.

OR13: I also wonder if switchin to YAML at the same time, might make this easier to work with.

mprorock: > I also wonder if switching to YAML at the same time, might make this easier to work with.

you know, YAML is not the worst idea in the world for this - might be worth exploring separately

OR13:
We need approvals or change requests on this PR.. please focus your comments on the content, we can address the JSON Schema encoding issues seperatly.

OR13 (changes requested):
continue the coding discussion in another PR, keep this one simple.

nissimsan:
Closing this, starting from a fresh PR just focusing on the ITI schema: https://github.com/w3c-ccg/traceability-vocab/pull/187

PR #180, opened 2021-07-05
[Chore] Readme cleanups

### LINKED ISSUE #174: Many Undefined Fields in published Spec
The Published Spec has many Undefined variables bubbling up to the surface: 

Not sure what's causing this one:
image

however these are due to missing descriptions:
image
image

Proposal: Have tests break on missing descriptions.

mkhraisha:
* fixes #174 by simply skipping description fields when generating the template for vocab.
this does not actually validate if the schema actually has a description field ( do we need that as a separate issue?)
* fixes Subtitle/Unofficial designation of the spec.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -11,7 +11,7 @@
     <script type="text/javascript" class="remove">
       var respecConfig = {
         // specification status (e.g., WD, LCWD, NOTE, etc.). If in doubt use ED.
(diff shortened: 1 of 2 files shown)


OR13 (approved):

mprorock (approved):

OR13:
We agreed to merge this on the call.

PR #182, opened 2021-07-09
Context Bug Fix

### LINKED ISSUE #150: Example Generation Adding @Context
In several spots in the Traceability Vocab, I've noticed that examples are being generated with contexts embedded into the credential subject and other properties. These additional contexts are not needed and we should fix the example generator to not display them. See example below for `credentialSubject `and `relatedDocuments`:

``` "credentialSubject": {
    "@context": [
      "https://w3id.org/traceability/v1"
    ],
    "type": [
      "BillOfLading"
    ],
    "billOfLadingNumber": "991205182",
    "relatedDocuments": [
      {
        "@context": [
          "https://w3id.org/traceability/v1"
        ],
        "type": [
          "Purchase"
        ],
        "customer": {
          "type": [
            "Person"
          ],
          "email": "Willie78@example.net",
          "phoneNumber": "555-152-7258"
        }
      }
    ],

### PR DESCRIPTION
**Pull request by mkhraisha**
fixes #150 

This PR currently strips all context objects, collects all unique ones in a set, then reattaches the set to the VC at root.

Issues: 
* There is a current issue that breaks on many credentials, I am unable to isolate why, I'm guessing something with order of contexts? 
to reproduce issue: 
modify ``` packages/traceability-schemas/ index.js``` to only target BillOfLading( works) or BillOfLadingCertificate(fails)
run ```npm run build:credentials```

### CODE REVIEW

**Comment by mkhraisha** [comment]
@OR13  would love it if you can take a look and help debug the ```@context``` issue

**Comment by TallTed** [comment]
Does this PR take into account @msporny's comments at https://github.com/w3c-ccg/traceability-vocab/issues/150#issuecomment-874303230?  Perhaps those may have some connection to the failure you're hitting?

I'm particularly wondering if your code has a way to handle multiple `@context` instances which map the same term differently at different levels.

**Comment by mkhraisha** [comment]
> Does this PR take into account @msporny's comments at [#150 (comment)](https://github.com/w3c-ccg/traceability-vocab/issues/150#issuecomment-874303230)? Perhaps those may have some connection to the failure you're hitting?
> 
> I'm particularly wondering if your code has a way to handle multiple `@context` instances which map the same term differently at different levels.

Manu's comment is about polluting the global namespace and the collision issues it may provide. 
the test cases that are failing are ones that I purposely chose would not run into this issue in order to get an initial simple solution that we can then test against an expanded data set.
All the nested objects in the BillOfLading and BillOfLadingCredential use the same two `@contexts`  throughout all nested objects: 
```
* 'https://www.w3.org/2018/credentials/v1'
* 'https://w3id.org/traceability/v1' 
```

They should not be running into any conflicting terms, and the order of the `@contexts` is preserved.

**Comment by OR13** [comment]
I said, I would look at this, and I did not. I will take a pass at figuring it out.

**Comment by OR13** [comment]
@nissimsan and I will take a look at this.

**Comment by nissimsan** [comment]
I'll action this for next time

**Comment by nissimsan** [comment]
@mkhraisha , VCs must contain @context which had gotten stripped out by the time it got to buildcredentials. 
https://github.com/Mavennet/traceability-vocab/pull/3/files
This forces a @context in a brute way and makes it work. Should probably be solved more elegantly.

**Comment by Therecanbeonlyone1969** [comment]
@mkhraisha @nissimsan @OR13 I looked at the code and the suggested solution https://github.com/Mavennet/traceability-vocab/pull/3/files. A thought here, rather than reinserting the first `@context` after stripping it, would it not be better to skip what we want to keep and not remove it like below?

```
function findAllByKey(obj, keyToFind) {
  return Object.entries(obj)
    .reduce((acc, [key, value]) => {
      if (key === keyToFind && value != '[
        'https://www.w3.org/2018/credentials/v1',
        'https://w3id.org/traceability/v1',
      ]') {
        contextArray.add(...value);
        // eslint-disable-next-line
        delete obj[key];
        return acc.concat(value);
      }
      if ((typeof value === 'object')) {
          return acc.concat(findAllByKey(value, keyToFind));
      }
      return acc;
      }, []);
}
```

**Comment by mkhraisha** [comment]
I've investigated further and found two separate issues: 
I was incorrectly adding the members to the set, and some Credentials (namely all that end in Certificate) where expecting an `@context` property even in the credentialSubject (as it is of type verifiableCredential) it seems to me theres a misconstruction going on there. 
To be clear the current BillOfLadingCertificate  is being wrapped by a VC instead of being treated like its own VC this is what is the root cause of the issue:

```
{
  '@context': [
    'https://www.w3.org/2018/credentials/v1',
    'https://w3id.org/traceability/v1'
  ],
  id: 'http://example.org/credentials/',
  type: [ 'VerifiableCredential' ],
  issuanceDate: '2021-02-04T20:29:37+00:00',
  issuer: 'did:key:z6MktHQo3fRRohk44dsbE76CuiTpBmyMWq2VVjvV6aBSeE3U',
  credentialSubject: {
    id: 'https://example.com/credential/123',
    type: [ 'VerifiableCredential', 'BillOfLadingCertificate' ],
    name: 'Bill Of Lading Certificate',
    description: 'This document includes recommended bill of lading fields.',
    relatedLink: [ [Object], [Object] ],
    issuanceDate: '2019-12-11T03:50:55Z',
    issuer: {
      id: 'did:example:123',
      type: 'Organization',
      name: 'Reichel Group',
      description: 'Total composite architecture',
      address: [Object],
      email: 'Morton33@example.com',
      phoneNumber: '555-718-6037',
      faxNumber: '555-183-5746'
    },
    credentialSubject: {
      type: [Array],
      billOfLadingNumber: '991205182',
      relatedDocuments: [Array],
      freight: [Object]
    },
    '@context': [
      'https://www.w3.org/2018/credentials/v1',
      'https://w3id.org/traceability/v1'
    ]
  }
}
```

mkhraisha:
On 17/8 call discussed this issue and decided to add a check on the type of object being issued and depending on if it is a VC or Subject then take appropriate action (wrap or issue)

PR #187, opened 2021-07-28
Intent To Import

nissimsan:
Removed commodity from the ITI as agreed on yesterday's call. 

Replaces https://github.com/w3c-ccg/traceability-vocab/pull/172


OR13 (approved):

mprorock (approved):

packages/traceability-schemas/schemas/IntentToImport.json
```diff
@@ -0,0 +1,31 @@
...
+            "oneOf": [
+                {
+                    "type": "string"
+                },
+                {
+                    "type": "array"
+                }
+            ]
+        },
+        "importer": {
```
mkhraisha, line 22: I may be missing something but doesn't this need to also reference a specific product? 
As far as I could tell all this is saying is "This organization intends to import", it does not reference where it is importing or what is it importing?

mprorock: [Intent To Sell](https://w3c-ccg.github.io/traceability-vocab/#IntentToSell) I think actually provides this capability with product definition (if available) - perhaps we rename that or merge these two objects - @OR13 thoughts?

OR13: I think the idea is to put that in other credentials presented along with this one.

mprorock (changes requested):
Consider merging with (or renaming) [IntentToSell](https://w3c-ccg.github.io/traceability-vocab/#IntentToSell)

OR13:
@mprorock this gets to the heart of accounting for why things move across boarders... sometimes its just another hop in the network, sometimes it's the destination... if it's the destination, trade finance / tariffs must be applied.

If "IntentToSell" really means "IntentToImport" and "I acknowledge the relevant financial implications of communicating this", then the semantics are aligned. 

I think it's also could be a mistake to conflate the product, and buyer and seller details  with the importers's acceptance of legal obligations... but I could be convinced.

TallTed:
This is one of those places where existing paper forms should be used to inform the new electronic structures ... and it should quickly become clear whether IntentToImport (perhaps for personal use — perhaps even personal property! — which may or may not trigger duties, tariffs, etc.) is indeed distinct from IntentToSell (which would generally trigger tariffs, among other things).

As may be clear from my phrasing above, I believe ITI is distinct from ITS.

That said, the same paper form might be used for both (and for other things), with different checkboxes and fields/sections to fill in, depending on the immediate activity.

packages/traceability-schemas/schemas/IntentToImport.json
nissimsan: Indeed, the idea was to have the ITI bundled with other credentials. I had an earlier PR with an ITI which included commodity code, but we decided to take that out. Which leads to the question: do we even need an ITI at all?

mprorock:
One of the things we need to define clearly is an intention vs an actual import in process.  When i defined IntentToSell originally, it was to specific that some party intended to sell an item across some geographic or physical border, so that that information could be conveyed in a verifiable manner to stakeholders to that possible future transaction.  Intent to Import appears to be getting at that same root issue, e.g. ability to notify some 3rd party that you intend to import something and and possibly put it on the market - this implies a few things:
1) that you define who you are
2) that you define who you are planning on importing from
3) that you define what you planning on importing
4) that you define when you are planning on importing it
and possibly:
5) what is the status of this, planning, actually in the process of importing it, changed your mind, etc

The exact scenario i had in mind when starting this out, is a case where a retailer notifies their own countries customs agency that they plan on importing a certain product (and if known a specified quantity of that certain product) as soon as they are aware that they might do so.  e.g. in our case, if a retailer request that a contract grower plants a certain crop in another country, with plans to import the harvested product, they could notify the appropriate officials as soon as they make that request - I am open to changing the term IntentToSell to IntentToImport as that is more reflective of the broader use case as @TallTed has noted (e.g. you may be just bringing something in without planning on actually selling that item) - note, this is different from a "declaration" form on actual import - as things change - this is to allow a verifiable way to notify early in the process that something may be crossing to border to potentially speed up a future import.

TallTed:
I'm still focusing on "how is this done today, with paper forms?"

Inventing an additional labor/time/money/something-saving electron-based device will be great, but IMHO should only come after the current processes are well understood and implemented by simply translating existing paper to electrons -- which is likely to save labor/time/money/something all by itself -- and won't require nearly the same kind of buy-in by users, as they can *easily* go from clipboards to tablets/phones/laptops with the same forms, while adding/re-ordering steps or forms in their process can disrupt everything....

mprorock:
> I'm still focusing on "how is this done today, with paper forms?"

Totally get it.  The declaration side, we absolutely have paper forms to model from that we need to closely align with.  

In this case, in regards to early notification, there is no existing process, and we are actually helping to define that early notification process, hence the need for a new object to capture the information

nissimsan:
@mprorock, what you describe sounds like a per-shipment, individually submitted [Advance Manifest](https://www.cbp.gov/bulletins/37genno52.pdf), but for the purpose of pre-processing normal customs procedures (not anti-terrorism). 

The attributes you list is a subset of a traditional [Customs Declaration](https://www.cbp.gov/sites/default/files/assets/documents/2019-Dec/CBP%20Form%207501_0.pdf) (except perhaps from `status`). 

I'm curious about the exact value which such earlier indication of this is aiming to achieve? And also where these requirements trace back to (which I may have missed still being relatively new in this group) - is this an expressed requirement from a customs authority? Or is it "just" a good idea ("look, customs, going from paper to digital presents new opportunities of slicing data up - it doesn't all have to fit on one piece of paper"). There's nothing wrong with the latter, as long as we treat it accordingly.

It would be a mistake for our mission to make ourselves dependant on changes to global customs processes. The essential purpose of this project is supporting the to move from paper to digital. Doing that and at the same time pushing for all shippers to include an extra "intention" step in their customs procedures would make our goal much, much harder to achieve.

mprorock:
There is definite overlap for sure.  This comes much earlier in the process
for us (e.g. potentially as early as before a crop is even planted).  Let's
discuss on the next call.

nissimsan:
Continuing on https://github.com/w3c-ccg/traceability-vocab/issues/194

PR #190, opened 2021-08-03
minor grammar and markup fixes in README.md

TallTed:


OR13 (approved):

README.md
```diff
@@ -61,7 +61,7 @@ need [`Node.js`](https://nodejs.org/en/) and [`git`](https://git-scm.com/).
 
 After you have the dependencies, the first-time setup is as follows:
 
-1. checkout this repository —
+1. check out this repository —
```
mkhraisha, line 64: both forms checkout vs check out are grammatically correct, however in this context we should be using checkout

README.md
```diff
@@ -291,11 +291,23 @@ npm run test:schemas
...
-3. OPTIONAL -- Add your new schema name to the file [includingevidence.json](./packages/traceability-schemas/src/data/vc/includingevidence.json) if you want to include the ```evidence``` property in your verifiable credential test fixture.
-4. OPTIONAL -- Add your new schema name to the file [includincredentialstatus.json](./packages/traceability-schemas/src/data/vc/includingcredentialstatus.json) if you want to include the ```credentialStatus``` property in your verifiable credential test fixture.   
-5. Run the build using: `npm run build:all`
-6. Fix any errors found.  Repeat as needed.
-7. Review the latest spec changes by serving docs: `npx serve ./docs`.
+3. OPTIONAL — If you want to include the ```evidence``` property in your
+   verifiable credential test fixture, add your new schema name to the
+   [includingevidence.json](./packages/traceability-schemas/src/data/vc/includingevidence.json)
+   file.
+5. OPTIONAL — If you want to include the ```credentialStatus``` property in
```
mkhraisha, line 298: I think you missed 4.

mkhraisha (changes requested):

README.md
TallTed: It's MD. I could use `1.` throughout or any assortment of numbers in the MD source; it'll go to clients as an HTML ordered list with ascending numerals.

README.md
TallTed: [Well](https://www.grammarphobia.com/blog/2019/05/checkout.html)...
>If the verb is intended, then the screen should read “Check Out,”  as if the instruction were short for “Check Out Here.”
>
>If the adjective is intended, the screen should read “Checkout”—as if short for “Checkout Option.”
>
>And if the noun is intended, the screen should also read “Checkout”—as if short for “Book Checkout.”

[Also](http://www.enhancemywriting.com/checkout-vs-check-out) --
> Checkout is a noun, and like most other nouns, it is only a single word. Check out is a phrasal verb, and like most other phrasal verbs it is two words.

Just to mess with us (as with many commands on Unix-like systems, which do not behave according to English rules), the `git` *command* is **`checkout`** -- which you use to **check out** the code.

mprorock (approved):

mkhraisha (approved):

PR #197, opened 2021-08-24
Bug/inject context fixtures

### LINKED ISSUE #150: Example Generation Adding @Context

mkhraisha:
fixes #150

This PR currently strips all context objects, collects all unique ones in a set, then reattaches the set to the VC at root.

Had to kill the old PR #182  as it was before the fixtures fix and had tons of thrashing.

### DIFF
--- a/packages/traceability-schemas/scripts/build-credentials.js
+++ b/packages/traceability-schemas/scripts/build-credentials.js
@@ -32,7 +32,6 @@ function findAllByKey(obj, keyToFind) {
       }, []);
 }
 
-
 const issueCreds = async (credTemplate, schemaName) => {
   try {
     const key = Ed25519KeyPair.from(require('../src/data/vc/keypair.json'));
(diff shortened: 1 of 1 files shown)


OR13 (approved):

mprorock (approved):
awesome!

PR #211, opened 2021-10-20
Add vocab for undefined terms

OR13:
Schema.org does the same thing, so I think its a good idea...

See also: https://github.com/w3c/vc-data-model/issues/753

Here is a poc showing the potential problems with this approach: 

https://github.com/acarnagey/verify-credential-test

Notice this unfortunate typo:

https://github.com/acarnagey/verify-credential-test/blob/865a1af971efa8f696b0e36bf9fe4b1611632a1a/documentLoader.js

In this case, the developer thought they were using the traceability context, but they were not.... all terms were being dropped.

If credentials v1 had an `@vocab`, this would not be possible.


nissimsan:
I got convinced to support this at last week's meeting.

mkhraisha (approved):

Issue #207, opened 2021-10-14
CD pipeline error


nissimsan:
The CD pipeline choked on my recent PR: 
https://github.com/w3c-ccg/traceability-vocab/actions/runs/1308699511

`main` renders fine locally: 
![image](https://user-images.githubusercontent.com/34443212/137301936-ee532233-29f7-4463-b696-0135e010b8f9.png)

OR13:
cc @mkhraisha  @mprorock  i suspect this is maybe related to the versioning feature.

OR13:
Closing, the pipeline is working.

PR #212, opened 2021-10-26
Upgrade node 16 + fix CD + npm ci

### LINKED ISSUE #202: Upgrade Node Version
We're currently using Node 12 ( at least according to CI) which is currently in maintenance LTS.
We should migrate to at least Node 14, I advocate we go to 16 though.

image

mkhraisha:
This PR fixes #202 #207 and modifies our `npm install` to `npm ci` 

This PR will not work as is because of the issues outlined in #207 , we need to create a service user to push to protected branches.

Changes:
* Modify `npm i` `npm install` to `npm ci`. This allows us to install from `package.lock` rather than resolving dependencies on the fly.
* Adds a requirement for `ACTION_AUTH_TOKEN ` to be defined in the secrets of the repo, which should correspond to a PAT for a user with admin privileges that can bypass the protected branch issue. [community](https://github.community/t/how-to-push-to-protected-branches-in-a-github-action/16101/4h)
* Migrates to Node 16 from Node 12.

DO NOT MERGE WITHOUT SETTING ACTION_AUTH_TOKEN


.github/workflows/cd.yml
```diff
@@ -34,13 +36,14 @@ jobs:
         uses: 'phips28/gh-action-bump-version@master'
         with:
           tag-prefix: 'v'
+          commit-message: 'CI: bumps version to {{version}} [skip ci]'
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.ACTION_AUTH_TOKEN }}
```
OR13, line 41: why is the necssary? can we just use this? 

https://github.com/decentralized-identity/JWS-Test-Suite/blob/main/.github/workflows/ci.yml#L32

We should not need secrets for this IMO, the default github tokens should be enough.

mkhraisha: The default github tokens cannot push to protected branches

.github/workflows/cd.yml
```diff
@@ -34,13 +36,14 @@ jobs:
...
 
       - name: Deploy test report
         uses: peaceiris/actions-gh-pages@v3
         with:
-          github_token: ${{ secrets.GITHUB_TOKEN }}
+          github_token: ${{ secrets.ACTION_AUTH_TOKEN }}
```
OR13, line 46: please use the secret available to all actions, we don't have branch protection on, so it should be sufficient.

mkhraisha: Will update!

mkhraisha:
Need to update the CI to be used on Push and on PR

mkhraisha:
This should fix the CI issues and it shoudl be ready now.

mprorock (approved):

mprorock:
@OR13 can you review

PR #222, opened 2021-11-30
Dcsa transport doc updates

nissimsan:
Various updates on the DCSA transport document and shipping instruction

PR #224, opened 2021-12-01
fest: dcsa transport document and shipping instructions

nissimsan:
Replacement of https://github.com/w3c-ccg/traceability-vocab/pull/222


nissimsan:
I've added an example based on this real BL sample. I'd like if we can discuss a good way to include such samples, close to their corresponding data representation. 
bill-of-lading-ex-1

OR13 (approved):

PR #226, opened 2021-12-07
Packing list updates

nissimsan:
Refactored the Packing List based on this sample, added as example. 

packing-list-ex


TallTed:
That is one mighty hand-crafted sample, only separated from hand-written by the use of Excel or other spreadsheet, *probably* primarily done to eliminate problems of illegible hand-writing!  Eliminating that sort of thing should definitely be a goal of this process...

In the meantime -- it would probably be useful to collect as many pre-printed forms as possible, distinctly from this sort of hand-crafted blank-page fake-form, and focus first on covering those pre-printed forms, which the hand-crafted versions are already imitating.

Otherwise, we're formalizing informal versions of formality, and risk omitting significant formalities which are already forgotten, distorted, or otherwise contaminated by the manual process.

nissimsan:
It's reality, @TallTed - this is the world we live in!  :)

mprorock (approved):
Cleans up a lot of cruft and standardizes in a way that can be used cross market sector

OR13 (approved):

TallTed:
"by ACRE AGE IS THE NEW BLACK" in the first column!  :-)

Issue #213, opened 2021-10-26
Don't invent new DateTime types


OR13:
originally from https://github.com/w3c-ccg/traceability-vocab/pull/210/files#r736753663

all instances of:

```
\"@type\": \"https://w3id.org/traceability#DateTime\"
```

should be like https://github.com/w3c/vc-data-model/blob/main/contexts/credentials/v1#L44:

```
\"@type\": \"http://www.w3.org/2001/XMLSchema#dateTime\"
```

rhofvendahl:
Addressed in https://github.com/w3c-ccg/traceability-vocab/pull/228

cc: @OR13

OR13:
cc @nissimsan

nissimsan:
Excellent input - many thanks @VladimirAlexiev. On it...!

OR13:
We did it.

PR #228, opened 2021-12-13
chore: replace w3id.org #DateTime types with w3.org #dateTime types

rhofvendahl:
Replaces all 6 instances of `\"@type\": \"https://w3id.org/traceability#DateTime\"` with `\"@type\": \"http://www.w3.org/2001/XMLSchema#dateTime\"`, as specified in https://github.com/w3c-ccg/traceability-vocab/issues/213.

### DIFF
--- a/packages/traceability-schemas/schemas/UsdaSc6.json
+++ b/packages/traceability-schemas/schemas/UsdaSc6.json
@@ -50,13 +50,13 @@
             "type": "string"
         },
         "dateOfEntry": {
-            "$comment": "{\"term\": \"dateOfEntry\", \"@id\": \"https://w3id.org/traceability#dateOfEntry\", \"@type\": \"https://w3id.org/traceability#DateTime\"}",
+            "$comment": "{\"term\": \"dateOfEntry\", \"@id\": \"https://w3id.org/traceability#dateOfEntry\", \"@type\": \"http://www.w3.org/2001/XMLSchema#dateTime\"}",
             "title": "dateOfEntry",
             "description": "Date when the the package entered the port of entry.",
             "type": "string"
         },
         "signatureDate": {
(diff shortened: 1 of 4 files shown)


packages/traceability-schemas/schemas/UsdaSc6.json
```diff
@@ -50,13 +50,13 @@
...
         "dateOfEntry": {
-            "$comment": "{\"term\": \"dateOfEntry\", \"@id\": \"https://w3id.org/traceability#dateOfEntry\", \"@type\": \"https://w3id.org/traceability#DateTime\"}",
+            "$comment": "{\"term\": \"dateOfEntry\", \"@id\": \"https://w3id.org/traceability#dateOfEntry\", \"@type\": \"http://www.w3.org/2001/XMLSchema#dateTime\"}",
             "title": "dateOfEntry",
             "description": "Date when the the package entered the port of entry.",
             "type": "string"
         },
         "signatureDate": {
-            "$comment": "{\"term\": \"signatureDate\", \"@id\": \"https://w3id.org/traceability#signatureDate\", \"@type\": \"https://w3id.org/traceability#DateTime\"}",
+            "$comment": "{\"term\": \"signatureDate\", \"@id\": \"https://w3id.org/traceability#signatureDate\", \"@type\": \"http://www.w3.org/2001/XMLSchema#dateTime\"}",
```
OR13, line 59: I question why we have this here... and why its not the `issuanceDate` or `proof.created` values... this is why Credential Types are needed not just JSON-LD Class types... still its orthogonal to the purpose of this PR.

OR13 (approved):
I think these changes are correct.

packages/traceability-schemas/schemas/UsdaSc6.json
mkhraisha: My understanding is this is a 1-1 from a physical document, I could be wrong though

mkhraisha (approved):

mprorock (approved):

Issue #239, opened 2021-12-24
Add example of usmcaProductSpecifier


OR13:

nissimsan:
```
{
  "id": "http://example.org/credentials/",
  "type": [
    "VerifiableCredential"
  ],
  "issuanceDate": "2021-02-04T20:29:37+00:00",
  "issuer": "did:key:z6MktHQo3fRRohk44dsbE76CuiTpBmyMWq2VVjvV6aBSeE3U",
  "credentialSubject": {
    "type": "UsmcaProductSpecifier",
    "product": {
      "type": [
        "Product"
      ],
      "sku": "491073039002",
      "description": "Non-alloy steel rolls",
      "commodity": {
        "type": [
          "Commodity"
        ],
        "commodityCode": "721320",
        "commodityCodeType": "HS",
        "description": "Steel Coils"
      }
    },
    "originCriterion": "A",
    "countryOfOrigin": "MX"
  },
  "@context": [
    "https://www.w3.org/2018/credentials/v1",
    "https://w3id.org/traceability/v1"
  ],
  "proof": {
    "type": "Ed25519Signature2018",
    "created": "2019-12-11T03:50:55Z",
    "jws": "eyJhbGciOiJFZERTQSIsImI2NCI6ZmFsc2UsImNyaXQiOlsiYjY0Il19..-umCHNa4s9eahgqvxDfcav3zQLguQydLYY2dXV_pZAnDZvNhgT5IFmGNpUfFdkAlY5lH9eRF-p2leDSkloU9Cg",
    "proofPurpose": "assertionMethod",
    "verificationMethod": "did:key:z6MktHQo3fRRohk44dsbE76CuiTpBmyMWq2VVjvV6aBSeE3U#z6MktHQo3fRRohk44dsbE76CuiTpBmyMWq2VVjvV6aBSeE3U"
  }
}
```

nissimsan:
https://github.com/w3c-ccg/traceability-vocab/pull/264

OR13:
The link to the schema is 404, still needs to be fixed: https://w3c-ccg.github.io/traceability-vocab/#USMCAProductSpecifier

BenjaminMoe:
![Screenshot 2022-08-31 at 02-50-50 Traceability Vocabulary v0 0](https://user-images.githubusercontent.com/86194145/187508137-ec79bb6d-c2d1-46f8-8de8-c55eb055fb17.png)

PR #252, opened 2022-01-05
chore: added usmca product specifier example

nissimsan:
https://github.com/w3c-ccg/traceability-vocab/issues/239


OR13 (approved):
This example makes sense, the data looks reasonable

mprorock (approved):

PR #259, opened 2022-01-21
Confirm CI

OR13:


OR13:
I merged this because somehow a broken example got into main without CI catching it.... and also this will make it much easier to verify examples signed by any did key.

PR #262, opened 2022-01-24
CTPAT feedback

nissimsan:
- Added extra elements based on conversations with CTPAT
- Switched to `@vocab` until `@context` is updated with the terms


OR13:
I don't think you sould need this:

```
{
        "@vocab": "https://w3id.org/traceability/v1/vocab#"
      }
```

since trace v1 already includes it, but seeing at it will be removed after this PR, I am fine leaving it in.

docs/openapi/components/schemas/common/CTPATCertificate.yml
```diff
@@ -16,31 +16,111 @@ description: >-
...
-    "type": "CTPATCertificate",
-    "ctpatCertifiedEntity": {
+    "type": [
+      "VerifiableCredential",
+      "CTPATCertificate"
+    ],
+    "@context": [
+      "https://www.w3.org/2018/credentials/v1",
+      {
+        "@vocab": "https://w3id.org/traceability/v1/vocab#"
```
OR13, line 107: does not appear to match the schema above, but we know we will fix this, later.

nissimsan: No, this is to solve the chicken-egg situation involved with introducing new terms now. It has to be a two-PR process (1. define the new terms, once published 2. use those new terms) - quite cumbersome. A tool for short circuiting this process seems essential for productivity.

nissimsan:
> I don't think you sould need this:
> 
> ```
> {
>         "@vocab": "https://w3id.org/traceability/v1/vocab#"
>       }
> ```
> 
> since trace v1 already includes it, but seeing at it will be removed after this PR, I am fine leaving it in.

... I also didn't think it would be necessary, but it is. Just confirmed it with another schema I was working on. It didn't verify in the local test (although it was a legit VC). But worked after adding the vocab. 

In any case, keeping this eyesore is a good incentive to switch to the proper terms when they're available on the context.

PR #269, opened 2022-01-25
Freight manifest schema

nissimsan:


mkhraisha:
What is the distinction between this freight manifest and https://github.com/w3c-ccg/traceability-vocab/blob/ab92c1341b1d361984b524736bfb5b4451c4cbfd/docs/openapi/components/schemas/common/BillOfLading.yml#L103

nissimsan:
@mkhraisha, a manifest relates to an entire vessel or airplane (not individual consignment) at a given time of entry or departure. It's information about all the cargo on board. For large vessels they can get quite big.

You might have seen passenger manifests getting printed at the gate in airports. It's that, just for cargo.

mkhraisha:
So the distinction then is 
Freight Manifest is the entire vessel's manifest
Freight property on product is for this individual product
is that accurate?

mprorock (approved):

mkhraisha (approved):

Issue #253, opened 2022-01-06
Redundant country attributes on PostalAddress


nissimsan:
Is there a good reason for having two country definitions on PostalAddress? They seem redundant. 

```
  addressCountry:
    title: Address Country
    description: >-
      Code specifying the country (and country subdivision) for the address
      using ISO 3166-1.
    type: string
    $linkedData:
      term: addressCountry
      '@id': https://schema.org/addressCountry
```

```
  countyCode:
    title: County Code
    description: >-
      A code that identifies a county. A county is a territorial division in
      some countries, forming the chief unit of local administration. In the US,
      a county is a political and administrative division of a state.
    type: string
    $linkedData:
      term: countyCode
      '@id': https://gs1.org/voc/countyCode
```

Fak3:
The one is countRy the other is county, they are different

nissimsan:
@Fak3 , 
> The one is countRy the other is county, they are different
... I don't understand your point. So you're saying both are needed? When would you use one over the other? Or both at the same time? 

PS: Super happy to see you here, @Fak3!! :)

Fak3:
> ... I don't understand your point. So you're saying both are needed? When would you use one over the other? Or both at the same time? 

A county is a territorial division in
      some countries, forming the chief unit of local administration.

TallTed:
(Surprisingly enough, dereferencing the URIs found in the `'@id'` attribute of the `$linkedData` object -- <https://schema.org/addressCountry> and <https://gs1.org/voc/countyCode> -- reveals descriptions which I think are sufficient to answer the question.)

nissimsan:
NOW I understand... Also what you meant by countRy, @Fak3. 
Thanks both - closing this.  :)

VladimirAlexiev:
@nissimsan you still need to fix the documentation:
```
 addressCountry:
    description: >-
      Code specifying the country (and country subdivision) for the address using ISO 3166-1.
```
You prolly got that description from https://www.gs1.org/voc/addressCountry.
But that prop has some schizofrenia coming from its GDSN/EDIFACT legacy: is it a country or a country subdivision?
See https://github.com/gs1/WebVoc/issues/25

https://schema.org/addressCountry has no such schizofrenia:
- "The country. For example, USA. You can also provide the two-letter ISO 3166-1 alpha-2 country code."
- Which you'd be justified to make more definite: "The two-letter ISO 3166-1 alpha-2 country code."

[...]

PR #282, opened 2022-01-26
feat: improved country code definition

nissimsan:
Ref: 
https://github.com/w3c-ccg/traceability-vocab/issues/253

### DIFF
--- a/docs/openapi/components/schemas/common/PostalAddress.yml
+++ b/docs/openapi/components/schemas/common/PostalAddress.yml
@@ -49,8 +49,7 @@ properties:
   addressCountry:
     title: Address Country
     description: >-
-      Code specifying the country (and country subdivision) for the address
-      using ISO 3166-1.
+      The two-letter ISO 3166-1 alpha-2 country code.
     type: string
     $linkedData:
       term: addressCountry


VladimirAlexiev (approved):

docs/openapi/components/schemas/common/PostalAddress.yml
```diff
@@ -49,8 +49,7 @@ properties:
   addressCountry:
     title: Address Country
     description: >-
-      Code specifying the country (and country subdivision) for the address
-      using ISO 3166-1.
+      The two-letter ISO 3166-1 alpha-2 country code.
```
mkhraisha, line 52: the schema.org entry for this has this text:
The country. For example, USA. You can also provide the two-letter ISO 3166-1 alpha-2 country code.
is there a reason why we are overloading it?

VladimirAlexiev: Because we want Traceability to be more definite than schema.org

mkhraisha: The change here seems super ripe for confusion, as all we're doing is modifying the description, while still linking to the term as defined by addressCountry.
In principle im absolutely all for being more definite but this appears ripe for confusion.

TallTed: I suggest that for our definition, we indicate the ISO code is preferred, but other designations (such as "USA" or "United States") are acceptable.  I suggest that this modified definition also be submitted as a PR against schema.org (upon merge of which, we can drop our distinct definition/property, and just use `schema:addressCountry`).

OR13:
Changes to be requested.

docs/openapi/components/schemas/common/PostalAddress.yml
VladimirAlexiev: I don't see what is confusing. IMHO it's fine to reuse external props but tighten them to what is more appropriate in your domain
This is called Application Profiles. It can be formalized in JSON Schema, SHACL or SHEX.

I disagree that Traceability should be more permissive since its domain (trade data exchange) is more tightly controlled than schema.org (eating data from billions of sites).
Eg if you don't use the ISO code, you cannot link to UNCEFACT top-level LOCODEs.

nissimsan: I completely agree with @VladimirAlexiev on this and find it difficult to actively "go backwards" and loosen up this definition. 

`The two-letter ISO 3166-1 alpha-2 country code.` is about as clear and concise as it can get, and I disagree that it is "super ripe for confusion". 

Let's lead (schema.org) by example here.

mprorock (approved):

docs/openapi/components/schemas/common/PostalAddress.yml
mkhraisha: I'm still not a fan, but I won't block this.

PR #283, opened 2022-01-27
fix: date type

nissimsan:
https://github.com/w3c-ccg/traceability-vocab/issues/213#issuecomment-1021340062

### DIFF
--- a/docs/openapi/components/schemas/common/IATAAirWaybillCertificate.yml
+++ b/docs/openapi/components/schemas/common/IATAAirWaybillCertificate.yml
@@ -171,8 +171,8 @@ properties:
     type: string
     $linkedData:
       term: requestedDate
-       '@id': https://w3id.org/traceability#requestDate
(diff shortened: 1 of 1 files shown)


mprorock (approved):

docs/openapi/components/schemas/common/IATAAirWaybillCertificate.yml
```diff
@@ -171,7 +171,8 @@ properties:
     type: string
     $linkedData:
       term: requestedDate
-      '@id': https://schema.org/Date
+       '@id': https://w3id.org/traceability#requestDate
+      '@type': http://www.w3.org/2001/XMLSchema#dateTime
```
OR13, line 175:
```suggestion
       '@type': http://www.w3.org/2001/XMLSchema#dateTime     
```

OR13 (changes requested):
Please fix whitespacing.

OR13 (approved):

OR13 (approved):

PR #303, opened 2022-02-03
Certificate of origin

nissimsan:


docs/openapi/components/schemas/common/CertificateOfOriginCertificate.yml
```diff
@@ -0,0 +1,134 @@
...
+          "type": "PostalAddress",
+          "streetAddress": "3934 Spinnaker Lane",
+          "addressLocality": "Joliet",
+          "addressRegion": "Illinois",
+          "postalCode": "60432",
+          "addressCountry": "US"
+        }
+      },
+      "carrier": {
+        "type": "Organization",
```
mkhraisha, line 90: This seems off, looks like you have an organization type that has manufacturing country and total number of packages on it?
im guessing this is a mistake in the generation of the example

mkhraisha (changes requested):
is the intention here that a CertificateOfOriginCertificate is just a consignment represented as a VC?
why not just call it ConsignmentCertificate?

mprorock (approved):
we need this as well.  thanks!

docs/openapi/components/schemas/common/CertificateOfOriginCertificate.yml
nissimsan: Ah yes! Definitely a mistake... Good catch.

nissimsan:
@mkhraisha, CertificateOfOrigin is an established type of document, no one would understand what a ConsignmentCertificate is. 
Admittedly CertificateOfOriginCertificate is a bit clunky, but opted for sticking hard to the convention which, as I understand it, postfixes "Certificate". 

@mprorock, yes, this is big in Agri: Wine from France, Maple syrup from Canada, etc...

mkhraisha:
Yes but then that should wrap an actual certificate thats purpose is to state the origin of the document, rather than wrapping a consignment. 
I agree that we need a certificate of origin but I disagree on that it should be just a wrapper around consignment

docs/openapi/components/schemas/common/CertificateOfOriginCertificate.yml
```diff
@@ -0,0 +1,134 @@
...
+  '@context':
+    type: array
+    const:
+      - 'https://www.w3.org/2018/credentials/v1'
+      - 'https://w3id.org/traceability/v1'
+  type:
+    type: array
+    const:
+      - VerifiableCredential
+      - CertificateOfOriginCertificatez
```
OR13, line 17: spelling.

OR13 (changes requested):
Lets find a better name for this.

mkhraisha:
in my mind a CertificateOfOrigin says: Product X was manufactured at Location Y by Entity Z at Time T, which while a consignment carries some/most of is a distinct thing from a consignment.

oldskeptic:
Agree with @nissimsan, while there are a few different flavours, a Certificate of Origin is the name commonly used in shipping for a specific type of document. I feel it should be attached to the consignment or shipment rather than wrapping it.

Fak3:
> Yes but then that should wrap an actual certificate thats purpose is to state the origin of the document, rather than wrapping a consignment. I agree that we need a certificate of origin but I disagree on that it should be just a wrapper around consignment

It seems more convenient to have the minimized credential data for origin of the product. It will enable to selectively disclose that piece of information without revealing other info that usually comes in the paper version of Certificate of Origin.
Multiple such VCs for a set of Products can be combined with Consignment VC into single Verifiable Presentation, and this presentation can be called ConsignmentCertificateOfOrigin, which is close to the paper version of this document

onthebreeze:
First there are three types of origin claim

- standard (non-preferential) certificates - issued by a chamber of commerce and attesting to origin of goods - use for trade finance and anti-dumping compliance
- preferential certificates - issued by the exporting customs authority or their accredited delegate (also a chamber of commerce in many cases).  explicitly related to the terms of a free trade agreement and allows the importer to claim preferential duty rates.
- declaration of origin - also used to colaim preferential duty rates but is a self-assessment issued by the exporter rather than an independent assessment by a third party such as a chamber. only some FTAs allow these self-declarations.  

all are much the same structure - except the preferential ones must reference the specific FTA and specific origin criteria specified within that FTA.

as for the shipment / consigment relationship. I thing the "subject" of an origin claim VC is the shipment (ie the commercial order of goods he goods) - they usually need to reference the commercial invoice.  Any reference to consignment (ie the transport order) is really just for the importing authority to know which box to look at.  so it's information on the CoO but not really the subject of the CoO

OR13:
@nissimsan can we find a way to tackle this PR in chunks? How can we move this forward given the feedback?

TallTed:
> can we find a way to tackle this PR in chunks? How can we move this forward given the feedback?

Merge this as a starting point, and shift the feedback into issues against that starting point, for which new PRs can then be developed.

mkhraisha:
My suggestion is to break this into two PRs:
One for consignment ( which I will happily accept)
and One for Certificate of Origin (which in this current state I am not at all ok)

nissimsan:
@onthebreeze, the way I read your requirements it's really the same schema we need for all three types of CoOs. The difference is only _who's the issuer_. 

I've attached some examples, which should make it clear why I took the approach I did: packaging, commodity, gross weight, marks and numbers, transport means, route, vessel nb, port of loading/discharge, container type... All these are screaming consignment. So I find `"credentialSubject": {"type": ["Consignment"]` a pretty elegant model, but @mkhraisha this is what you are reacting against, right? Admittedly, it is somewhat UN/CEFACT-centric, though, and I'm okay to change this. Piece count and net weight are also arguably at a finer grained invoice-level. 

Here's a suggestion: I'm currently also working on some commercial invoice improvements, particularly adding line items. How about making the CoO's subject an array of such [TradeLineItems](https://github.com/w3c-ccg/traceability-vocab/blob/ci-updates/docs/openapi/components/schemas/common/TradeLineItem.yml)?
@mkhraisha if you agree with this, then I'll be happy to make the change (I might just wait until after the TradeLineItem is merged to avoid git gymnastics). 

image
image

nissimsan:
@mkhraisha, quickly recapping what we've discussed offline: not all types of CoOs focuses consignment/transport. For example NAFTA and USMCA does not. And neither does the ERGaR gas CoO you shared. 

Looking at @onthebreeze's [list above](https://github.com/w3c-ccg/traceability-vocab/pull/303#issuecomment-1033624776), there seems to be a pattern that non-preferential CoOs focuses on consignment aspects, preferential CoOs focuses on other targeted elements. 

In this case, I propose that we rename this one to 'NonPreferentialCertificateOfOrigin` and add the dedicated preferential CoOs separately: `USMCACertificateOfOrigin` (which we have already), `ERGaRCertificateOfOrigin`, and so on... 

@mkhraisha, @onthebreeze, does that sound appropriate?

mkhraisha:
Preferential vs non-preferential is not the boundary that delineates the difference between those two, there are non-preferential COs that do not contain consignment data, and there are preferential that do.

I believe the correct approach is to have a CO that has many optional fields including fields of a consignment.
So to be clear I am proposing that we should have:

1. A Consignment Schema
2. A Consignment Certificate
3. A Certificate of Origin Schema
4. A Certificate of Origin Certificate

Specific Preferential COs that track a physical document can and should be added the same way we're tracking existing documents.

onthebreeze:
An origin claim is always linked to a specific shipment - at the goods (shipment) level not the transport (consignment) level. The key reference on every origin claim is the commercial invoice number that represents the shipment of goods. All must have that because the verifier (typically a trade finance institution and/or an importing customs authority needs to correlate the origin claim to a physical commercial deal.  

The reason you sometimes do and sometimes don't see consignment references on origin claims is because that information is only needed for pre-clearance activities of importing regulators - ie "which container should I open if I want to physically confirm that the shipped goods correspond to what is claimed on the document.  

There are three types of origin claim:
- non preferential certificates of origin. Issued by chambers and always linked to goods on an invoice.  Used for trade finance and anti-dumping and not part of importing regulator pre-clearance activities so often don't have a consignment reference - not don't confuse that with the idea that it's sone kind of multi-shipment multi-use certificate. It is not.
- preferential certificate of origin. Issued either by an exporting customs authority or their authorised delegate.  It's purpose is to attest to the specific origin criteria that corresponds to a "preferential" tariff under the terms of a specific free trade agreement.  Basically it's for the importer to pay less duty.  It can also serve the same purpose as non-preferential so you'd never issue both for one shipment.  Much more frequently has a consignment reference so that importing customs know what box to open. When a pref cert does not have a consignment reference on the document it's only because it's superfluous for the importing regulator because they have other ways to know which box to open (eg the CoO is attached to an import declaration which itself references a cargo movement that has the very sane consignment reference )
- a declaration of origin (DoO) serves the same purpose as a preferential certificate of origin except that it is issued as a self-assessed declaration by the exporter and not by a competent authority. Some FTA allow DoOs and some don't.  Sone FTA allow them but only in the context of an "advanced ruling" from the importing authority. A DoO may or may not include a consignment t reference for the same reason a pref CoO might - it's to do with importing authority capability 

Finally there is no such thing as a "consignment certificate".  An AWB or BoL issued as a VC would be the closest thing.  

So what does all this mean in the verifiable linked data space? I would suggest that you strongly resist the temptation to model your digital world on the paper world. That doesn't mean that you aren't "paper compatible" - this is a separate implementation discussion about how you digitise a paper centric supply chain in a scalable incremental way that doesn't demand every stakeholder is digital ready.  What I mean about avoiding paper modelling is that you should understand why something is on paper and ask what the linked data equivalent is. 

All three origin certificate types can be modelled the same way. They are all an origin claim (aka vocabulary uri) made by an issuer (did) about one or more trade items (uri) in a shipment (uri of invoice or did of shipment, depending on your implementation choice). Optionally also the uri/did of the consignment document. 

So basically just three bits of data.  Everything else on an origin certificate is derivable from those linked data identifiers

nissimsan:
@mkhraisha, I changed the credentialSubject now from Consignment to product TradeLineItems. These are listed on CIs too ("uri of invoice or did of shipment" gets too abstract for my taste).

@onthebreeze , I agree completely, ConsignmentCertificate doesn't make sense (that's a Bill of Lading!)

docs/openapi/components/schemas/common/CertificateOfOrigin.yml
```diff
@@ -0,0 +1,180 @@
...
+  totalNumberOfPackages: 
+    title: Total Number of Packages
+    description: A number of packages.
+    type: number
+    $linkedData:
+      term: totalNumberOfPackages
+      '@id': https://service.unece.org/trade/uncefact/vocabulary/uncefact/#packageQuantity
+  dateOfExport:
+    title: Date of Export
+    description: The date, time, date time or other date time value when this supply chain consignment will exit, or has exited from the last port, airport, or border post of the country of export.
```
TallTed, line 93:
```suggestion
    description: The date, time, date time, or other date time value when this supply chain consignment will exit, or has exited from the last port, airport, or border post of the country of export.
```

mkhraisha:
@nissimsan I appreciate that you have updated the credential subject to be line items, that makes sense to me. 
However the certificate still uses consignment language everywhere, for example it requires a "consignee", "shipper" and others on the base of the certificate. 
It should IMO instead be an option al field that links to a consignment VC. 
Sorry for being difficult about this, but if we follow that flow then Certificates of Origin similar to this one: [Certificate of Origin 1](https://www.atlastube.com/wp-content/uploads/2018/02/2018-NAFTA-Certificate-of-Origin.pdf)
or [Certificate of Origin 2](https://www.ergar.org/wp-content/uploads/2018/10/210705_ERGaR-CoO-Scheme-Rules-v1.1.pdf), won't be modeled well.

I think it should the absolute bare minimum that we can then extend out to USMCACertificateOfOrigin and others. 
The absolute minimum to me are Manufacture Location, Manufacturer, and dates.

OR13 (approved):

mkhraisha:
To clarify the CR currently is: the certificate still uses consignment language everywhere, for example it requires a "consignee", "shipper" and others on the base of the certificate.
I would like this language to be updated.

Happy to have this merged if we raise an inheritance issue to have a base Certificate of Origin, with extensions for USMCA etc.

nissimsan:
@mkhraisha , have a look now - I've removed all those consignment-related parties. It's pretty stripped down now, I'd say it works well as a generic super class in this shape.

docs/openapi/components/schemas/common/CertificateOfOrigin.yml
```diff
@@ -0,0 +1,94 @@
...
+  manufacturingCountry: 
+    title: Manufacturing Country
+    description: Manufacturing country.
+    type: string
+    $linkedData:
+      term: countryOfOrigin
+      '@id': https://w3id.org/traceability#countryOfOrigin
+  dateOfExport:
+    title: Date of Export
+    description: The date, time, date time, or other date time value when this supply chain consignment will exit, or has exited from the last port, airport, or border post of the country of export.
```
mkhraisha, line 47:
```suggestion
    description: The date, time, date time, or other date time value when this supply chain consignment will exit, or has exited from the last port, airport, or border post of the country of export.
```
```
description: The date, time, date time, or other date time value when the subject line items will exit, or has exited from the last port, airport, or border post of the country of export.
```

mkhraisha:
minor comment, PR good to merge though!

docs/openapi/components/schemas/common/CertificateOfOrigin.yml
TallTed, line 47: Turns @mkhraisha's suggestion into a suggestion, and fixes a plurality disagreement.
```suggestion
    description: The date, time, date time, or other date time value when the subject line item(s) will exit, or has(have) exited from the last port, airport, or border post of the country of export.
```

mkhraisha (approved):
Thank you for all the hard work!

PR #316, opened 2022-02-26
fix: bugs

nissimsan:
Syntax bugs and @vocab deletion

### DIFF
--- a/docs/openapi/components/schemas/common/FreightManifestCertificate.yml
+++ b/docs/openapi/components/schemas/common/FreightManifestCertificate.yml
@@ -34,9 +34,7 @@ example: |-
   {
     "@context": [
       "https://www.w3.org/2018/credentials/v1",
-      {
-        "@vocab": "https://w3id.org/traceability/v1/vocab#"
-      }
+      "https://w3id.org/traceability/v1"
     ],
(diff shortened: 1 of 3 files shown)


rhofvendahl:
@nissimsan 

I only saw this after creating https://github.com/w3c-ccg/traceability-vocab/pull/317, & it looks like you've addressed a few of the same bugs. Once this is merged I'll update my PR so it doesn't conflict with these changes.

mkhraisha (approved):
Looks great!

nissimsan:
@rhofvendahl - thank you!! 

I was forced to introduce those `@vocab`s in the interim period before we had [this tool](https://github.com/w3c-ccg/traceability-vocab/tree/main/packages/traceability-schemas/data). It's been bugging me ever since, causing annoying side effects (you might have noticed). Typical that we just decided to pick that up at essentially the same time! :D

Thanks again! If you're on the call tomorrow we'll get to it then too.

TallTed:
@nissimsan -- Please edit the initial comment on this PR, to codefence `@vocab` (yes, another github user we needlessly `@` ping when writing too quickly...)

OR13 (approved):

PR #317, opened 2022-02-27
fix: clean up openapi yaml files

rhofvendahl:
This PR cleans up a number of minor issues with the YAML schema files.

Of these changes, a few are pretty much guesses and definitely need review:
- There were a handful of examples where additional properties were passed to entities with `additionalProperties: false` (often "ID" to Organization instances). I chose to remove the additional parameters being passed in.
  - in CTPATCertificate.yml
  - in DCSATransportDocument.yml
  - in USMCACertifier.yml
- Several example `@context`s listed something like:
  ```
      "@context": [
        "https://www.w3.org/2018/credentials/v1",
        {
          "@vocab": "https://w3id.org/traceability/v1/vocab#"
        }
      ],
  ```
  which conflicted with `@context` defined as a const:
  ```
    '@context':
      type: array
      const:
        - 'https://www.w3.org/2018/credentials/v1'
        - 'https://w3id.org/traceability/v1'
  ```
  In these cases I changed the const to:
  ```
    const:
      - 'https://www.w3.org/2018/credentials/v1'
      - '@vocab': 'https://w3id.org/traceability/v1/vocab#'
  ```
- A few schemas were missing `$linkedData`. I filled these in with best guesses:
  - Context.yml
  - TraceablePresentation.yml
  - Workflow.yml
- In one case a `$linkedData` was similarly missing from a property:
  - SIMASteelImportLicenseApplication.yml (`productInformation` property)
- In Transport.yml there were two required properties that weren't actually present as properties, which I removed.

(Changes I'm more confident about not mentioned)

### DIFF
--- a/docs/openapi/components/schemas/common/CTPATCertificate.yml
+++ b/docs/openapi/components/schemas/common/CTPATCertificate.yml
@@ -103,9 +103,7 @@ example: |-
     ],
     "@context": [
       "https://www.w3.org/2018/credentials/v1",
-      {
-        "@vocab": "https://w3id.org/traceability/v1/vocab#"
-      }
+      "https://w3id.org/traceability/v1"
     ],
     "id": "did:key:z6LSpdSReUHCjYcQb1243aF1vS7sd9ArK585Mm4ktARQxatd",
     "name": "CTPAT Certificate",
@@ -115,12 +113,10 @@ example: |-
       "name": "United States Customs Trade Partnership Against Terrorism",
       "description": "Your Supply Chain's Strongest Link",
       "email": "ctpathelpdesk@cbp.dhs.gov",
-      "phoneNumber": "1-800-927-8729",
-      "id": "did:key:z6MkjArXVLoKSmFgBXNpug5eehrwpy9hLLbBq7rT2wYamdDr"
+      "phoneNumber": "1-800-927-8729"
     },
     "issuanceDate": "2022-01-13T09:16:46Z",
     "credentialSubject": {
-      "id": "did:key:z6MkhfZ7sNYEJ8vFSpwJaeyN7zNUaTxS4TBxW3y6R9ZKRaQ4",
       "type": "Organization",
       "name": "Trans-Atlantic Shipping Co. Ltd.",
       "address": {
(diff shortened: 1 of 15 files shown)


rhofvendahl:
Update: For some reason the "Verifiable Credentials" test in credentials.sanity.test.js only passes when I revert CTPATCertificate.yml to include IDs for issuer and credentialSubject. As ID is not a property defined in the Organization schema (which those properties reference) and Organization does not allow additional properties, this triggers a console.error message.

I'll leave the IDs in so the test passes, but if anyone has any suggestions I'd love to hear them.

nissimsan:
@rhofvendahl, again: thanks!! 

> In these cases I changed the const to:
> ```
>   const:
>     - 'https://www.w3.org/2018/credentials/v1'
>     - '@vocab': 'https://w3id.org/traceability/v1/vocab#'
> ```
... [As mentioned](https://github.com/w3c-ccg/traceability-vocab/pull/316#issuecomment-1054645050) these `@vocab`s shouldn't survive. 

Our fixes to TraceablePresentation.yml and Workflow.yml seems largely identical. 

I refactored the SIMA licences a fair bit today.

OR13 (changes requested):
the syntax is not correct, `@vocab` should probably not be added like this.

nissimsan:
@rhofvendahl , we still stick with the order discussed last week (and [here](https://github.com/w3c-ccg/traceability-vocab/pull/316#issuecomment-1053015906)), right: 
- First merge https://github.com/w3c-ccg/traceability-vocab/pull/319
- Then adjust [this](https://github.com/w3c-ccg/traceability-vocab/pull/317)

mkhraisha (approved):
Looks reasonable, but I haven't actually tested these. 
@OR13 would you be able to do a sanity check on this?

mprorock (approved):

mkhraisha:
`@vocab` pr has been merged, this needs conflict fixes and should be good to go.

OR13 (changes requested):
Conflicts and change requests need to be addressed,  suggest merging https://github.com/w3c-ccg/traceability-vocab/pull/355 before this one.

PR #323, opened 2022-03-01
Add Placeholder `PurchaseOrder` and `PurchaseOrderCertificate`

BenjaminMoe:
This pull request adds a `PurchaseOrderCertificate`, which is a verifiable credential wrapper for a `PurchaseOrder`.
Currently the `PurchaseOrder` is loosely typed allowing for credentials to be issued. 
Definitions for the `PurchaseOrder` schema will be added in subsequent pull requests.

### DIFF
--- a/docs/openapi/components/schemas/common/PurchaseOrder.yml
+++ b/docs/openapi/components/schemas/common/PurchaseOrder.yml
@@ -1,3 +1,6 @@
+$linkedData:
+  term: PurchaseOrder
+  '@id': https://w3id.org/traceability#PurchaseOrder
 title: PurchaseOrder
 type: object
 description: A statement issued by a buyer for the sale of products or services to be delivered at a later date
--- a/docs/openapi/components/schemas/common/PurchaseOrder.yml
+++ b/docs/openapi/components/schemas/common/PurchaseOrder.yml
@@ -1,0 +1,8 @@
+title: PurchaseOrder
+type: object
+description: A statement issued by a buyer for the sale of products or services to be delivered at a later date
+additionalProperties: true
+example: |-
+  {
+    "type": "PurchaseOrder"
+  }
--- a/docs/openapi/components/schemas/common/PurchaseOrderCertificate.yml
+++ b/docs/openapi/components/schemas/common/PurchaseOrderCertificate.yml
@@ -1,0 +1,82 @@
+$linkedData:
+  term: PurchaseOrderCertificate
+  '@id': https://w3id.org/traceability#PurchaseOrderCertificate
+title: Commercial Invoice Certificate
+description: Certifications made about a purchase order
+type: object
+properties:
+  '@context':
+    type: array
+    const:
+      - 'https://www.w3.org/2018/credentials/v1'
+      - 'https://w3id.org/traceability/v1'
+  type:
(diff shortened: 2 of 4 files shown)


OR13 (approved):

nissimsan (changes requested):
@BenjaminMoe, wonderful that you are adding documentation to the "future tool" on the /data/README.md. But (despite using Purchase Order as the example), that part really has nothing to do with this PR. 
I suggest you split that into a separate PR.

BenjaminMoe:
@nissimsan Thanks the feedback. I was thinking that separation of concerns could be handled better here. I'll leave this PR up as adding placeholders for the Purchase Order documents, and split the documentation updates into a separate PR.

mkhraisha:
How is a purchase order significantly different from the existing [Purchase Schema](https://github.com/w3c-ccg/traceability-vocab/blob/main/docs/openapi/components/schemas/common/Purchase.yml)? 
Looks like a Purchase wraps a Purchase order, do we want to have both in the vocabulary?

BenjaminMoe:
A Purchase Order is a pretty common B2B document, where a buyer is requesting goods to be delivered at a certain place, at a certain time, fitting certain conditions. 

For example, in the case of just in time manufacturing, an OEM would order parts from several suppliers on the conditions they are delivered ready to be assembled. 

The Purchase schema, seems to refer to a `purchaseOrder` in the definition. I'm guessing the string is for a PurchaseOrder No., but no example is provided. 

```
  purchaseOrder:
    title: Purchase Order
    description: The purchase order associated with a purchase
    type: string
    $linkedData:
      term: purchaseOrder
      '@id': https://schema.org/identifier
 ```
 
As for PurchaseOrderCertificate, versus PurchaseOrder. This is for conventions. A PurchaseOrder is the document for the good to be purchased, the delivery date, and any other conditions required for payment. A PurchaseOrderCerticiate is verifiable credential signed by a buyer.

BenjaminMoe:
> @BenjaminMoe, wonderful that you are adding documentation to the "future tool" on the /data/README.md. But (despite using Purchase Order as the example), that part really has nothing to do with this PR. I suggest you split that into a separate PR.

Documentation moved to: https://github.com/w3c-ccg/traceability-vocab/pull/333/files

mkhraisha:
Are you advocating for both a Purchase Schema and a Purchase Order schema?
If so we probably need a clear line of delineation between them, especially since a Purchase has a purchase order identifier on it. 
I was working under the assumption that a Purchase **is** a Purchase Order, but I may be missing something.

if these are distinct entities, then it may be a good idea to link Purchase Order id from Purchase, which should be in a different PR, just flagging it as a potential issue to be raised.

BenjaminMoe:
For clarification, yes. A Purchase **is not a** Purchase Order. 
A Purchase is a transaction.
A Purchase Order is a request for goods to be delivered on a certain date, under certain conditions.

I make a Purchase Order to have a number of clothing items delivered to a store. The delivery doesn't make it on time. Thus no Purchase takes place.

mkhraisha (approved):

mprorock (approved):

nissimsan (approved):

PR #335, opened 2022-03-08
feat: example data for HBL, MBL and SI

nissimsan:
Added examples which was left blank in the first iteration. 
Also changed from common `Consignment` subject to dedicated schemas, to define the different parts of consignment needed.

### DIFF
--- a/docs/openapi/components/schemas/common/HouseBillOfLadingCertificate.yml
+++ b/docs/openapi/components/schemas/common/HouseBillOfLadingCertificate.yml
@@ -4,7 +4,7 @@ $linkedData:
 title: House Bill of Lading Certificate
 description: >- 
   A bill of lading issued by a freight forwarder. Often covers a consignment of parcels from various shippers that has been grouped or consolidated by the forwarder. The forwarder may, for example, receive a single groupage bill of lading from the carrier, then issue a series of House B/Ls to the respective shipper.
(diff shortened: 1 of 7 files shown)


BenjaminMoe (approved):

docs/openapi/components/schemas/common/HouseBillOfLadingCertificate.yml
```diff
@@ -38,43 +38,102 @@ properties:
     items:
       $ref: ./LinkRole.yml
 additionalProperties: false
-required: []
 example: |-
```
mkhraisha, line 41: The example attached here is for a masterBOLcert not a houseBOL cert

docs/openapi/components/schemas/common/MasterBillOfLadingCertificate.yml
```diff
@@ -66,15 +60,76 @@ example: |-
       }
     },
     "credentialSubject": {
-      "type": [
-        "Consignment"
+      "type": "Consignment",
```
mkhraisha, line 63: Shouldn't this Type be MasterBOL?

mkhraisha (changes requested):

docs/openapi/components/schemas/common/MasterBillOfLadingCertificate.yml
nissimsan: Yes!

mprorock (approved):

PR #339, opened 2022-03-10
Add PGAStatusMessage and PGAStatusMessageCertificate

BenjaminMoe:
Add `PGAStatusMessage` and `PGAStatusMessageCertificate`.

### DIFF
--- a/docs/openapi/components/schemas/common/PGAStatusMessageCertificate.yml
+++ b/docs/openapi/components/schemas/common/PGAStatusMessageCertificate.yml
@@ -98,9 +98,9 @@ example: |-
     ],
     "proof": {
       "type": "Ed25519Signature2018",
-      "created": "2022-03-14T08:56:39Z",
+      "created": "2022-03-14T13:41:24Z",
       "verificationMethod": "did:key:z6MktHQo3fRRohk44dsbE76CuiTpBmyMWq2VVjvV6aBSeE3U#z6MktHQo3fRRohk44dsbE76CuiTpBmyMWq2VVjvV6aBSeE3U",
       "proofPurpose": "assertionMethod",
-      "jws": "eyJhbGciOiJFZERTQSIsImI2NCI6ZmFsc2UsImNyaXQiOlsiYjY0Il19..Z24zJafaE54R1wUBK5nTNrkXoFo0JjXVgWEe461-8dWDCsYvsi207GrIQW79RMQn0bPDWwzpV_1v64jOrgWqCw"
+      "jws": "eyJhbGciOiJFZERTQSIsImI2NCI6ZmFsc2UsImNyaXQiOlsiYjY0Il19..RQiFHEwYQgPhZd4exxvMytJZZG6lsgndwSxi4v2h_KbhO0Q_-W2L8E_XVJAHc50NDIU5_LYO_SXIiIE1276SAw"
     }
-  } 
+  }
--- a/docs/openapi/components/schemas/common/PGAStatusMessageCertificate.yml
+++ b/docs/openapi/components/schemas/common/PGAStatusMessageCertificate.yml
@@ -51,7 +51,7 @@ example: |-
     "id": "https://example.com/credential/123",
     "type": [
       "VerifiableCredential",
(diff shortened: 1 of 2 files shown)


docs/openapi/components/schemas/common/PGAStatusMessageCertificate.yml
```diff
@@ -0,0 +1,110 @@
...
+      "PGAStatusMessageCertificate"
+    ],
+    "name": "PGA Hold Certificate",
+    "description": "Certificate Issued by a PGA for the status of Cargo",
+    "relatedLink": [],
+    "issuanceDate": "2019-12-11T03:50:55Z",
+    "issuer": {
+      "id": "did:key:z6MktHQo3fRRohk44dsbE76CuiTpBmyMWq2VVjvV6aBSeE3U",
+      "type": "Organization",
+      "name": "Waters Inc",
```
nissimsan, line 63: @BenjaminMoe , I recommend you make this issuer more relatable. "Waters Inc" does not sound at all like a PGA. 
Call it something like "Food Border Force Agency" and fbf.example.gov.

docs/openapi/components/schemas/common/PGAStatusMessageCertificate.yml
```diff
@@ -0,0 +1,110 @@
...
+        "entryNo": "AAA-1234567-8",
+        "entryLineSequence": "L1:S1",
+        "statusCode": "O2",
+        "statusCodeDescription": "Hold Intact",
+        "validCodeReason": "25",
+        "validCodeReasonDescription": "Additional Verification Needed",
+        "subReasonCode": "132",
+        "subReasonCodeDescription": "Documentation Needed"
+      },
+      {
```
nissimsan, line 90: Is this to be understood as the second message which is sent about the same shipment? And the first message is just kept around, but would have been sent already? 
Typically this would include a timestamp of some sort. (But I don't want to invent non-existing requirements).

docs/openapi/components/schemas/common/PGAStatusMessageCertificate.yml
```diff
@@ -0,0 +1,110 @@
...
+        "recordNo": "SO70",
+        "entryNo": "AAA-1234567-8",
+        "entryLineSequence": "L1:S2",
+        "statusCode": "O7",
+        "statusCodeDescription": "May Proceed",
+        "validCodeReason": "23",
+        "validCodeReasonDescription": "Released",
+        "subReasonCode": "",
+        "subReasonCodeDescription": ""
+      }
```
nissimsan, line 101: This messaging scheme feels somewhat old fashioned. I'm sure this reflects how this is done today. I'd much rather use our Workflow for linking follow-up messages. 
The balance of course it to not take to many steps at once - this is probably a good first step.

nissimsan (changes requested):
Make the issuer more relatable. No action needed on my other comments.

docs/openapi/components/schemas/common/PGAStatusMessageCertificate.yml
BenjaminMoe: Thanks for the feedback. Will update with the provided PGA name.

docs/openapi/components/schemas/common/PGAStatusMessageCertificate.yml
BenjaminMoe: My guess for this scenario is that PGA's are likely going to be managing thousands of shipments. And each record is likely going to contain multiple line numbers. So it would make sense to allow for multiple messages for respective records and line numbers to be exchanged in a single credential.

docs/openapi/components/schemas/common/PGAStatusMessageCertificate.yml
BenjaminMoe: We can make these optional fields if they are optional. The main reason for including these is that I prefer not to deal with JSON that changes structure when it can be avoided. Makes it marginally easier to work with.

nissimsan (approved):
Beautiful!

OR13 (approved):

PR #375, opened 2022-04-13
Add Software Supply Chain to Respec Doc

nissimsan:


docs/index.html
```diff
@@ -416,6 +416,11 @@ <h2>Structure</h2>
               <td>Mahmoud Alkhraishi</td>
               <td>mahmoud@mavennet.com</td>
             </tr>
+            <tr>
+              <td>Software Bill of Materials</td>
```
BenjaminMoe, line 420: Should we call this "software supply chain"? A software bill of materials is one of the possible certificates that can be issued.

docs/index.html
```diff
@@ -696,6 +701,26 @@ <h5>Sample E-Commerce Workflow</h5>
 back to the Express Carrier (14) well in advance of the goods arrival.
         </p>
       </section>
+      <section>
+        <h5>Software Bill of Materials</h5>
```
BenjaminMoe, line 705: Similar issue here, the field is "software supply chain"

docs/index.html
```diff
@@ -696,6 +701,26 @@ <h5>Sample E-Commerce Workflow</h5>
...
+        <dl class="termlist">
+          <p>
+            Software is widely built from open source and other types of components. A Software Bill of Material
```
BenjaminMoe, line 708: We could start with the "why" as opposed to the "what" . The why is that physical supply chains now depend on software supply chains. Being able to secure digital services has become a requirement for securing physical ones. 

It's become a national security issue, and that's why the US president admin has issed an executive order. One of the approaches for increased visibility and accountability in software is to provide a software bill of materials which allows consumers to understand what's in the software the same way food ingredients are labeled.

OR13 (changes requested):
agree with @BenjaminMoe 's feedback.

docs/index.html
nissimsan: For sure!

docs/index.html
```diff
@@ -702,21 +702,30 @@ <h5>Sample E-Commerce Workflow</h5>
...
           </p>
-          <p>
-            The May 21st, 2021 US President <a
-              href="https://www.whitehouse.gov/briefing-room/presidential-actions/2021/05/12/executive-order-on-improving-the-nations-cybersecurity/">
-              Executive Order</a> of enhancing software supply chain security is a notable example of recent surging
-            software traceability requirements. As with physical supply chains, customers demand vendors to take
-            responsibility and prove the provenance of software products and platforms. Whether commercial or open source,
-            including the Software Bill of Material in a Verifiable Credential establishes a verifiable link from the
+          <p>            
+            Including the Software Bill of Material in a Verifiable Credential establishes a verifiable link from the
```
TallTed:
```suggestion
            Including the Software Bill of Materials in a Verifiable Credential establishes a verifiable link from the
```

BenjaminMoe (approved):

OR13 (approved):

PR #377, opened 2022-04-15
Update Organization & related examples

rhofvendahl:
This PR would  replace both the `globalLocationNumber` and `address` props of Organization with `place`.

Benefits:
- Place has its own `globalLocationNumber` and `address` props, so this would simplify Organization without any loss of information
- This would allow the association of additional location info (like geo coordinates) with organizations.

I was able to replace `address` with `place` in most examples, but ran into some difficulties doing this with certificates (CommercialInvoiceCertificate, DCSAShippingInstructionCertificate etc.). Even after adding a `place` prop to Organization, adding `place` props to example organizations in any of the certificates resulted in failed verification checks (from `packages/traceability-schemas/__tests__/credentials.sanity.test.js`).

As far as I can tell the verification checks _do_ check the credentials against the latest Organization schema (as removing `address` from Organization without removing all `address` props from certificate examples breaks them), so I'm really not sure why I'm not able to add `place` properties to certificate example organizations after adding `place` to the latest Organization schema. Is there maybe some cache of vocab related info being referenced by `credentials.sanity.test.js` that needs to be updated before I can update certificate example organizations? @OR13 maybe you could shine some light on this?

For now I've just removed `address` props for organizations in certificate examples (& updated the proofs).

### DIFF
--- a/docs/openapi/components/schemas/common/Organization.yml
+++ b/docs/openapi/components/schemas/common/Organization.yml
@@ -50,20 +50,6 @@ properties:
     $linkedData:
       term: description
       '@id': https://schema.org/description
-  globalLocationNumber:
-    title: Global Location Number
-    description: The GS1 GLN of the organization.
-    type: string
-    $linkedData:
-      term: globalLocationNumber
-      '@id': https://schema.org/globalLocationNumber
-  address:
-    title: Postal Address
-    description: The postal address for an organization or place.
-    $ref: ./PostalAddress.yml
-    $linkedData:
-      term: address
-      '@id': https://schema.org/PostalAddress
   email:
     title: Email Address
     description: Organization's primary email address.
@@ -115,6 +101,41 @@ properties:
     $linkedData:
       term: iataCarrierCode
       '@id': https://onerecord.iata.org/cargo/Company#airlineCode
+  organizationType:
+    title: Organization Type
+    description: Type of organization (e.g. cold storage, meat processing).
+    type: string
+    $linkedData:
+      term: organizationType
+      '@id': https://w3id.org/traceability#organizationType
+  place:
+    title: Place
+    description: Information on organization place.
+    $ref: ./Place.yml
+    $linkedData:
+      term: place
+      '@id': https://schema.org/Place
+  gap:
+    title: GAP Certifications
+    description: List of GAP certifications and links.
+    type: string
+    $linkedData:
+      term: gap
+      '@id': https://w3id.org/traceability#gap
+  generalCerts:
+    title: General Certifications
+    description: List of general certificates (such as export certificates, etc.) possessed by the organization.
+    type: string
+    $linkedData:
+      term: generalCerts
+      '@id': https://w3id.org/traceability#generalCerts
+  trustedBy:
+    title: Trusted By
+    description: List of VCs issued showing that the organization is a known and trusted entity.
+    type: string
+    $linkedData:
+      term: trustedBy
+      '@id': https://w3id.org/traceability#trustedBy
 additionalProperties: false
 example: |-
   {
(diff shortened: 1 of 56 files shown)


OR13 (approved):

mprorock (approved):

docs/openapi/components/schemas/common/Organization.yml
```diff
@@ -115,20 +101,72 @@ properties:
...
       term: iataCarrierCode
       '@id': https://onerecord.iata.org/cargo/Company#airlineCode
+  organizationType:
+    title: Organization Type
+    description: Type of organization (e.g. cold storage, meat processing).
+    type: string
+    $linkedData:
+      term: organizationType
+      '@id': https://w3id.org/traceability#organizationType
+  place:
```
nissimsan, line 111: @rhofvendahl, I love the ambition of normalizing our data model and applaud the effort in working through the cascading impacts! But replacing an attribute which is defined on schema.org (`Organization.address`) with one which isn't `Organization.place` doesn't seem right.
The attribute should be `Organization.location`, ref: https://schema.org/Organization

docs/openapi/components/schemas/common/Organization.yml
```diff
@@ -115,20 +101,72 @@ properties:
...
+    title: Place
+    description: Information on organization place.
+    $ref: ./Place.yml
+    $linkedData:
+      term: place
+      '@id': https://schema.org/Place
+  gap:
```
nissimsan, line 118: What's GAP? Please spell out acronyms.

nissimsan (changes requested):

docs/openapi/components/schemas/common/Organization.yml
mkhraisha: [Good Agricultural Practices (GAP)](https://www.ams.usda.gov/services/auditing/gap-ghp) but yes this should be in the description at a minimum

docs/openapi/components/schemas/common/Organization.yml
```diff
@@ -115,20 +101,72 @@ properties:
...
+      term: gap
+      '@id': https://w3id.org/traceability#gap
+  generalCerts:
+    title: General Certifications
+    description: List of general certificates (such as export certificates, etc.) possessed by the organization.
+    type: string
+    $linkedData:
+      term: generalCerts
+      '@id': https://w3id.org/traceability#generalCerts
+  trustedBy:
```
mkhraisha, line 132: How do we expect this to work?
is this a group of organizations that say they trust this organization? what happens if they no longer trust this organization? will this organization need to reissue its VC?

mkhraisha (changes requested):
Not too sure how the trusted by field will work and have some reservations outlined.

docs/openapi/components/schemas/common/Organization.yml
OR13: seems like a subtype specific to ag.

docs/openapi/components/schemas/common/Organization.yml
OR13: yeah, interested to understand this

OR13 (changes requested):
address merge conflicts

docs/openapi/components/schemas/common/Organization.yml
mprorock: @rhofvendahl  GAP will only apply to ag organizations - should not be a top level member of organization - GAP should be an ag inspection in reference to the organization

mprorock: @rhofvendahl you may want an AgOrganization that includes org as a top level member, then adds gap and other certs as required

docs/openapi/components/schemas/common/Organization.yml
mprorock: This will likely go another route - common scenario we encounter is an attestation by a 3rd party that they trust some org to have X internal credential - i think in our case AgInspection or another more general cert that references the org is better than embedding in the org.

@rhofvendahl i suggest removing generalCerts and trustedBy for now - and perhaps add a general attestation credential - e.g. org A says the following claims about org B

docs/openapi/components/schemas/common/Organization.yml
```diff
@@ -115,20 +101,72 @@ properties:
...
+    title: Trusted By
+    description: List of VCs issued showing that the organization is a known and trusted entity.
+    type: string
+    $linkedData:
+      term: trustedBy
+      '@id': https://w3id.org/traceability#trustedBy
```
mprorock, line 138:
```suggestion
```

docs/openapi/components/schemas/common/Organization.yml
mprorock, line 132:
```suggestion
```

docs/openapi/components/schemas/common/Organization.yml
mprorock, line 118:
```suggestion
```

docs/openapi/components/schemas/common/Organization.yml
```diff
@@ -115,20 +101,72 @@ properties:
...
+    $linkedData:
+      term: place
+      '@id': https://schema.org/Place
+  gap:
+    title: GAP Certifications
+    description: List of GAP certifications and links.
+    type: string
+    $linkedData:
+      term: gap
+      '@id': https://w3id.org/traceability#gap
```
mprorock, line 124:
```suggestion
```

docs/openapi/components/schemas/common/Organization.yml
```diff
@@ -115,20 +101,72 @@ properties:
...
+      term: generalCerts
+      '@id': https://w3id.org/traceability#generalCerts
+  trustedBy:
+    title: Trusted By
+    description: List of VCs issued showing that the organization is a known and trusted entity.
+    type: string
+    $linkedData:
+      term: trustedBy
+      '@id': https://w3id.org/traceability#trustedBy
 additionalProperties: false
```
mprorock, line 139:
```suggestion
additionalProperties: true
```

mprorock (changes requested):

TallTed:
@rhofvendahl -- Please make a comment indicating why you've withdrawn this PR...

PR #387, opened 2022-04-20
Find replace unsupported oneOf

OR13:
oneOf is not supported by our JSON-LD processor. 

This PR removes Entity which is basically a wrapper around oneOf person or organization.

IMO, this was a mistake from the beginning, since encoding and encouraging optionality destroys quality of the certificate.

Additionally, the unsupported oneOf is breaking integrations on our side, which we are hoping to share with the group related to inlining of schemas... this is necessary to fix those inlining issues.


TallTed:
(Where is the issue and associated discussion that leads to this PR?  This should have started as an issue, with time and space for relevant discussion leading to a planned fix, then leading to a PR such as this...)

> oneOf is not supported by our JSON-LD processor.

Is this `Entity` a legal JSON-LD construction, whether a wrapper or not?  I can think of many instances where there is only a person, not an organization -- which makes your suggested "fix" problematic.

I submit that the fix for the described issue should a patch to your JSON-LD processor, not removal of the `Entity` construct.

nissimsan:
https://github.com/w3c-ccg/traceability-vocab/pull/388 would this be an alternative way of making the processor pick `Organization` instead of `Person`.

BenjaminMoe:
Does this extend to other instances of `oneOf`? We have a lot of instances where `type` can either be a string, or an array of strings. Does this mean that we want to enforce a convention of always using an array of strings?

As for the proposed changes, I think we can use a work-around. My impression is that `Entity` is supposed to represent a legal entity which can be an organization or person. I don't know that we can unilaterally assert than an entity will always be an organization in every instance. 

Since `Organization` and `Person` share a lot of the same attributes, we could flatten `Entity` to have an `EntityType` property, that would be an enum for `Person` or `Organization`. And that would allow us to get the functionality of `oneOf` without using it.

OR13:
@BenjaminMoe this does apply to all instance of oneOf... we reported errors regarding this ages ago, the fixes to the problem are not worth the cost, and they are undermining the quality of the data... I don't think we should continue down this path, I think we should remove the use of oneOf... at least until its possible to handle it better, and I am not committed to doing that work based on our current roadmaps and timeline.

mprorock:
> This PR removes Entity which is basically a wrapper around oneOf person or organization.
> 
> IMO, this was a mistake from the beginning, since encoding and encouraging optionality destroys quality of the certificate.

Entity (as a person or an organization) is critical for agricultural use cases.  hard -1 on this.  Either we fix oneOf, or adjust approach

OR13:
@mprorock when can we expect a PR for your fixes? its non trivial to implement, and currently broken.

OR13:
I would suggest we merge this PR, so its not broken, and then you can take your time implementing an extension.

OR13:
conflicts exist, and @BenjaminMoe and @nissimsan are going to sort through removing any case where a credential we care about is blocked by this unsupported feature.

PR #388, opened 2022-04-21
Change order of Person and Org

nissimsan:
To make the LD processor pick Org instead of Person

### DIFF
--- a/docs/openapi/components/schemas/common/Entity.yml
+++ b/docs/openapi/components/schemas/common/Entity.yml
@@ -4,8 +4,8 @@ $linkedData:
 title: Entity
 description: A person or organization.
 oneOf:
-  - $ref: ./Person.yml
   - $ref: ./Organization.yml
+  - $ref: ./Person.yml
 example: |-
   {
     "type": "Person",


OR13 (changes requested):
This is a bad idea.

TallTed:
> This is a bad idea.

This "argument" is untouchable, because it says nothing.  Basis for the label "bad" is necessary to evaluate whether the PR should be merged or rejected.

nissimsan:
To be clear: what I'm suggesting here is a major hack exploiting that the LD processor picks the first of the oneOf list (in itself an assumption). This was meant as a suggestion to buy time, keeping the current `Entity` around while hoping that the  LD processor would magically evolve to support oneOf.

mprorock:
> To be clear: what I'm suggesting here is a major hack exploiting that the LD processor picks the first of the oneOf list (in itself an assumption). This was meant as a suggestion to buy time, keeping the current `Entity` around while hoping that the LD processor would magically evolve to support oneOf.

This seems like a good short term hack until someone fixes the LD processor

OR13:
This is not a good hack, it's not a solution to the problem.

TallTed:
> This is not a good hack, it's not a solution to the problem.

Closing this PR with the declaration above doesn't feel appropriate to me.  It feels like an arbitrary decision based on one person's sense of a rather indeterminate problem.

I don't think we have agreement on what that problem is, so no suggested (partial or complete) hack, workaround, nor other solution, is likely to be sufficient to the resolve that problem.

I submit that this PR, and all other PRs aimed at this still-vague problem, should be held open until discussion on the now-open issue (#395) leads to a succinct and complete description of the problem, regarding resolution of which each PR may then be evaluated -- hopefully leading to one PR that resolves the finally described problem, in favor of which the others may then be closed.

BenjaminMoe:
> I don't think we have agreement on what that problem is, so no suggested (partial or complete) hack, workaround, nor other solution, is likely to be sufficient to the resolve that problem.

@TallTed this is the reason we closed https://github.com/w3c-ccg/traceability-vocab/pull/387 and https://github.com/w3c-ccg/traceability-vocab/pull/388. 

In issue https://github.com/w3c-ccg/traceability-vocab/issues/395 we need to define the following before considering Pull Requests on this issue. 
- A clear definition of the problem
- Potential work a rounds or solutions and their pros and cons

If we deem that allowing the JSON-LD processor to fail is an acceptable solution, then we can re-open this Pull Request.

PR #391, opened 2022-04-21
Template schema

nissimsan:
Schema for representing our schemas as templates.

### DIFF
--- a/docs/openapi/components/schemas/common/Template.yml
+++ b/docs/openapi/components/schemas/common/Template.yml
@@ -80,10 +80,11 @@ example: |-
         "Freight Forwarder",
         "Carrier"
     ],
-    "schema": {},
     "tags": [
         "common"
     ],
+    "schema": {},
+    "uiSchema": {},
     "formData": {
         "@context": [
             "https://www.w3.org/2018/credentials/v1",
--- a/docs/openapi/components/schemas/common/Template.yml
+++ b/docs/openapi/components/schemas/common/Template.yml
@@ -1,0 +1,226 @@
+$linkedData:
+  term: Template
+  '@id': https://w3id.org/traceability#Template
+title: Template
+description: A template for rendering schemas for example with JSON Schema Forms.
+type: object
+properties:
+  type:
+    oneOf:
+      - type: array
+        items:
+          type: string
+          enum:
+            - Template
+      - type: string
+        const:
+          - Template 
+  id:
+    type: string
+  industryTags:
+    title: Industry Tags
+    description: Industry vertical applicability.
+    type: array
+    items:
+      type: string
+  industryRoleTags:
+    title: Industry Role Tags
+    description: Industry role applicability.
+    type: array
+    items:
+      type: string
+  tags:
+    title: Tags
+    description: Tags applicable to the template.
+    type: array
+    items:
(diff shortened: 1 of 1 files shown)


docs/openapi/components/schemas/common/Template.yml
```diff
@@ -0,0 +1,227 @@
...
+    title: Updated
+    description: Template update date.
+    type: string
+required:
+  - id
+  - industryTags
+  - industryRoleTags
+  - tags
+  - schema
+  - uiSchema
```
acarnagey, line 68: I think uiSchema shouldn't be required

acarnagey:
could add optional field type object xhrSchema

docs/openapi/components/schemas/common/Template.yml
OR13: agree... also wondering if we should derive it from the $linkedData approach or include it in parallel like this.

docs/openapi/components/schemas/common/Template.yml
```diff
@@ -0,0 +1,227 @@
...
+  - formData
+  - organization
+additionalProperties: false
+example: |-
```
acarnagey, line 72: I don't think you need the `|-`, it was making the whole example a string instead of object

docs/openapi/components/schemas/common/Template.yml
```diff
@@ -0,0 +1,227 @@
...
+    description: Layout definitions
+    type: object
+  formData:
+    title: Form Data
+    description: Placeholder example data.
+    type: object
+  organization:
+    title: Organization
+    description: The organization creating the template.
+    $ref: ./Organization.yml
```
acarnagey, line 53: don't need title and description since ref will have it, should be `$ref: './Organization.yml'`

docs/openapi/components/schemas/common/Template.yml
nissimsan: This is how we typically do it in trace-vocab. It's a way to allow multi-line example, which looks better in the bottom of the yaml files.

docs/openapi/components/schemas/common/Template.yml
nissimsan: This title and desc is for this attribute. Like, this particular usage of the Org class. That's not the same as the definition of the Org class. 

You couldn't add "The organization creating the template." in the definition of Organization.

nissimsan:
> could add optional field type object xhrSchema, also optional image field type string.

Can you elaborate, pls, @acarnagey? What would xhrSchema be used for? 

+1 on adding image.

acarnagey:
xhrSchema seems to center around things having to do with api calls
I think there are ways have same functionality without it, like using onSubmit, onChange callbacks. Having xhrSchema adds some more flexibility.
It includes functionality for things like:
- mapping form to request body when submitting form to api
- what api call to make on form submission
- what api call to make on form load to prepopulate fields
- what ui error on form to show when api fails
- for a specific drop down what api call to fetch options

nissimsan:
xhrSchema seems like a step too far, @acarnagey.

BenjaminMoe (approved):

mkhraisha (approved):

Issue #410, opened 2022-05-19
Ag Examples should be updated with better data


mprorock:
@rhofvendahl can you update all ag examples so that measurements are correct for ag and so that all example data makes sense.  Also please coordinate with me on explicit VC examples such as PPQ203 etc that are just in vocab now and should be in the "Credentials" section

PR #413, opened 2022-05-24
chore: update ag examples with realistic data

rhofvendahl:
Addresses https://github.com/w3c-ccg/traceability-vocab/issues/410

### DIFF
--- a/docs/openapi/components/schemas/common/ppq203.yml
+++ b/docs/openapi/components/schemas/common/ppq203.yml
@@ -114,7 +114,7 @@ example: |-
           "type": [
             "ChemicalProperty"
           ],
-          "name": "Cesium"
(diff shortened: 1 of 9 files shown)


mprorock (approved):

docs/openapi/components/schemas/common/ppq203.yml
```diff
@@ -131,16 +131,18 @@ example: |-
         "date": "2019-12-11T03:50:55Z",
         "property": {
           "type": [
-            "ChemicalProperty"
+            "MechanicalProperty"
```
mprorock, line 134: This will need a further update - e.g. should be more something like a contamination by white flies

nissimsan (approved):

PR #418, opened 2022-06-06
Realistic MTR sample and example

nissimsan:
- This adds a real sample of an MTR (logo, address and signature is fake company for redaction).
- Plus it updates the example correspondingly, test are based on that sample.


BenjaminMoe (approved):

OR13 (approved):

docs/openapi/components/schemas/common/MillTestReport.yml
```diff
@@ -58,30 +58,27 @@ example: |-
       "type": [
         "Organization"
       ],
-      "name": "Brekke, Kertzmann and Yost",
-      "description": "Distributed mobile protocol",
+      "id": "did:key:z6MkkDM7AG9KSX8naVHuD4oyx16FX4k2EwJmPrSUfrK5fhEu",
+      "name": "Melt Global, Inc.",
+      "phoneNumber": "+1 555-323-9000",
```
TallTed, line 63: Example phone numbers should be properly constructed, i.e., for US, using a REAL area code (which requires a REAL city/town and state/province, though optimally an obviously FAKE street address and company name, such as "fake co, 15 fake street, Houston, TX") with a RESERVED EXAMPLE local number (`555-01xx`).  The `555` area code is not [currently assigned](https://en.wikipedia.org/wiki/North_American_Numbering_Plan), but it is also not currently reserved -- so it could be assigned at any point in the future.

mkhraisha (approved):

docs/openapi/components/schemas/common/MillTestReport.yml
nissimsan: Thanks @TallTed! That was exactly what the 555 was meant to do - didn't know it was unreserved.

nissimsan: Done - updated to a phone nb from a faker service

TallTed: Fixed one problem, kept a similar one. `702-647-9292` may not ring through today, but `647-9292` is a valid local number, and `702` is a real Las Vegas area code.

`702-555-0101` would be fully fake, as `555-0101` falls within the reserved range of `555-0100`-`555-0199` (a/k/a `555-01xx`).  Do you need a PR or issue from me on this, or is this comment sufficient?

Issue #230, opened 2021-12-14
Specific vs. generic identifiers


nissimsan:
We recently discussed whether a BOL number is best described with the more general 
`https://schema.org/identifier`
or the more narrow/precise 
`https://service.unece.org/trade/uncefact/vocabulary/uncl1153/#Bill_of_lading_number`

Neither is wrong, a BL# is both of these definitions. I can see use cases and arguments for both, and I find it difficult to decide on the best approach without applying uninformed guesswork regarding the data consumer. 

I don't think the argument of "always choose the generic option" holds here. Choosing 
`https://schema.org/Organization`
over 
`https://service.unece.org/trade/uncefact/vocabulary/uncefact/#Organization`
is straight forward because it's the _exact same_ meaning, and many more will understand the former. For BL# the latter is a _type of_ the former.

nissimsan:
Would it make any sense to allow for multiple/array of term definitions?

TallTed:
A reasoner can be brought to bear such that an `https://service.unece.org/trade/uncefact/vocabulary/uncl1153/#Bill_of_lading_number` is _implicitly_ also an `https://schema.org/identifier`.  This is generally optimal.

However, a reasoner is not always usable, for various reasons.  In these cases, it would often if not always be better to multi-type the entity -- that is, an `https://service.unece.org/trade/uncefact/vocabulary/uncl1153/#Bill_of_lading_number` is *explicitly* also an `https://schema.org/identifier.

When starting from a specialized ontology of this sort, it is often helpful to include broaderThan, narrowerThan, equivalentProperty, equivalentClass, subClass, subProperty, and similar statements relating the more specialized ontology to the more generic.  These make it viable for following users to use whichever descriptor (a/k/a attribute, predicate, property, etc.) makes sense to them at the time -- that is, it doesn't matter whether they use `https://schema.org/Organization` or `https://service.unece.org/trade/uncefact/vocabulary/uncefact/#Organization` -- but it certainly will have an impact if they use the very broad and generic `https://schema.org/identifier` instead of the very narrow and specific `https://service.unece.org/trade/uncefact/vocabulary/uncl1153/#Bill_of_lading_number`.

VladimirAlexiev:
Despite the capitalization, `uncl1153:Bill_of_lading_number` is a property, just like `schema:identifier` is a property.
- @nissimsan: the JSONLD context must map a term `billOfLading` to one or the other, it cannot map it to both
[...]

PR #432, opened 2022-06-13
feat: create Ag Grade Inspection & supporting schemas

rhofvendahl:
This PR introduces a new `AgGradeInspection` schema, as well as a number of additional supporting schemas. A minor fix in `AgParcelDelivery` was necessary as well.

The original intention was to replace `AgInspectionReport`, but after investigating I believe it wouldn't be possible to create a general ag inspection form that encompasses all of the information necessary to describe ag grade inspections. I'll likely update `AgInspectionReport` as well as other ag inspection schemas in later PRs.

### DIFF
--- a/docs/openapi/components/schemas/common/AgGradeInspectionDefect.yml
+++ b/docs/openapi/components/schemas/common/AgGradeInspectionDefect.yml
@@ -1,0 +1,61 @@
+$linkedData:
+  term: AgGradeInspectionDefect
+  '@id': https://w3id.org/traceability#AgGradeInspectionDefect
+title: Agriculture Grade Inspection Defect
+description: Information on a type of offsize / defect observed.
+type: object
+properties:
+  type:
+    oneOf:
+      - type: array
+        items:
+          type: string
+          enum:
+            - AgGradeInspectionDefect
+      - type: string
+        const:
+          - AgGradeInspectionDefect
+  offsizeDefect:
+    title: Offsize / Defect
+    description: Brief description of this offsize / defect.
+    type: string
+    $linkedData:
+      term: offsizeDefect
+      '@id': https://schema.org/description
+  averageDefects:
+    title: Average Defects
+    description: Percent of sample items with defects for the grade standards.
+    type: number
+    $linkedData:
+      term: averageDefects
+      '@id': https://schema.org/Number
+  damage:
+    title: Percent Damage
+    description: Percent of sample items with damage (but not serious or very serious damage).
+    type: number
+    $linkedData:
+      term: damage
+      '@id': https://schema.org/Number
+  seriousDamage:
+    title: Serious Damage
+    description: Number of sample items with serious damage (incl. decay, soft rot, or soft).
+    type: number
+    $linkedData:
+      term: seriousDamage
+      '@id': https://schema.org/Number
+  verySeriousDamage:
+    title: Very Serious Damage
+    description: Number of sample items with very serious damage (incl. significant decay, soft rot, or soft).
+    type: number
+    $linkedData:
+      term: verySeriousDamage
+      '@id': https://schema.org/Number
+additionalProperties: false
+example: |-
+  {
+    "type": "AgGradeInspectionDefect",
+    "offsizeDefect": "Quality - Dry Sunken Areas",
+    "averageDefects": 2,
+    "seriousDamage": 0,
+    "verySeriousDamage": 0
+  }
--- a/docs/openapi/components/schemas/common/AgGradeInspectionLot.yml
+++ b/docs/openapi/components/schemas/common/AgGradeInspectionLot.yml
@@ -1,0 +1,255 @@
+$linkedData:
+  term: AgGradeInspectionLot
+  '@id': https://w3id.org/traceability#AgGradeInspectionLot
+title: Agriculture Grade Inspection Lot
+description: Information regarding the inspected lot including samples, defects and grades.
+type: object
+properties:
+  type:
+    oneOf:
+      - type: array
+        items:
+          type: string
+          enum:
+            - AgGradeInspectionLot
+      - type: string
+        const:
+          - AgInspectionLotesult
+  agProduct:
+    title: Product
+    description: The product inspected.
+    $ref: ./AgProduct.yml
+    $linkedData:
+      term: agProduct
+      '@id': https://w3id.org/traceability#AgProduct
+  lotIdentifier:
+    title: Lot Identifier
+    description: Use to further add to identification of the lot inspected, or to distinguish between lots / sublots for certification purposes (e.g. "Loaded", "Main Lot", "U.S. Fancy", "Film Wrapped").
+    type: string
+    $linkedData:
+      term: lotIdentifier
+      '@id': https://schema.org/identifier
+  numberContainers:
+    title: Number of Containers
+    description: Number of containers in the lot or portion of lot inspected, as well as type of package or container in which the product was shipped (e.g. "672 cartons", "20,000 LBS.").
+    type: number
+    $linkedData:
+      term: numberContainers
+      '@id': https://schema.org/Integer
+  countInspected:
+    title: Count Inspected
+    description: Whether the number of containers was accurately inspected. If false, report the reason for not being able to count the containers in the "Remarks" section.
+    type: boolean
+    $linkedData:
+      term: countInspected
+      '@id': https://w3id.org/traceability#countInspected
+  brandMarkings:
+    title: Brand / Markings
+    description: Brand or Markings on inspected product. ALways have brand name enclosed in quotation marks as first entry (e.g. "Unkle Okey's Onions"), with other markings as necessary to identify lot NOT enclosed in quotation marks (e.g. U.S. No. 1; or Grown & Packed by Unkle Okey, Inc.; or Produce of U.S.A.).
+    type: string
+    $linkedData:
+      term: brandMarkings
+      '@id': https://schema.org/description    
+  samples:
+    title: Samples
+    description: List of samples taken for this lot.
+    type: array
+    items:
+      $ref: ./AgGradeInspectionSample.yml
+    $linkedData:
+      term: sample
+      '@id': https://w3id.org/traceability#AgGradeInspectionSample
+  defects:
+    title: Defects
+    description: List of defects observed for this lot.
+    type: array
+    items:
+      $ref: ./AgGradeInspectionDefect.yml
+    $linkedData:
+      term: defect
+      '@id': https://w3id.org/traceability#AgGradeInspectionDefect
+  grade:
+    title: Grade
+    description: The grade assigned to this lot.
+    $ref: ./AgGradeInspectionResult.yml
+    $linkedData:
+      term: grade
+      '@id': https://w3id.org/traceability#AgGradeInspectionResult
+  remarks:
+    title: Remarks
+    description: Additional information, such as a description of the lot (e.g. size, color, quality). Generally includes factors which do not affect grade.
+    type: string
+    $linkedData:
+      term: remarks
+      '@id': https://schema.org/description
+  minTemperature:
+    title: Minimum Temperature
+    description: Minimum product temperature measured.
+    $ref: ./MeasuredValue.yml
+    $linkedData:
+      term: minTemperature
+      '@id': https://schema.org/MeasuredValue
+  maxTemperature:
+    title: Maximum Temperature
+    description: Maximum product temperature measured.
+    $ref: ./MeasuredValue.yml
+    $linkedData:
+      term: maxTemperature
+      '@id': https://schema.org/MeasuredValue
+  pli:
+    title: PLI
+    description: "Positive Lot Identification (PLI) is the application of identifying marks to an inspected lot. For more information see: https://www.ams.usda.gov/publications/content/positive-lot-identification-pli-manual-worksheets"
+    type: string
+    $linkedData:
+      term: pli
+      '@id': https://w3id.org/traceability#pli
+additionalProperties: false
+example: |-
+  {
+    "type": "AgGradeInspectionLot",
+    "agProduct": {
+      "type": [
+        "AgProduct"
+      ],
+      "upc": "033383401508",
+      "plu": "94225",
+      "gtin": "033383401508",
+      "product": {
+        "type": [
+          "Product"
+        ],
+        "manufacturer": {
+          "type": [
+            "Organization"
+          ],
+          "email": "Ashlee.Grady@example.net",
+          "phoneNumber": "555-899-1399"
+        },
+        "name": "Avocados",
+        "description": "Avocados, 4 pack boxes",
+        "sizeOrAmount": {
+          "type": [
+            "QuantitativeValue"
+          ],
+          "unitCode": "hg/ha",
+          "value": "60"
+        },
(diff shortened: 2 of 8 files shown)


docs/openapi/components/schemas/common/AgGradeInspectionDefect.yml
```diff
@@ -0,0 +1,61 @@
+$linkedData:
+  term: AgGradeInspectionDefect
```
OR13, line 2: I find the `Ag` abbreviation sorta annoying... from a convention perspective, I believe we should avoid abbreviations, as first time readers struggle with them... its an audience thing, not everyone will identify `Ag` as agriculture, especially non english speakers.

docs/openapi/components/schemas/common/AgGradeInspectionDefect.yml
```diff
@@ -0,0 +1,61 @@
...
+  type:
+    oneOf:
+      - type: array
+        items:
+          type: string
+          enum:
+            - AgGradeInspectionDefect
+      - type: string
+        const:
+          - AgGradeInspectionDefect
```
OR13, line 17: @BenjaminMoe is this the right syntax for this? I can never remember.

docs/openapi/components/schemas/common/AgGradeInspectionDefect.yml
```diff
@@ -0,0 +1,61 @@
...
+  offsizeDefect:
+    title: Offsize / Defect
+    description: Brief description of this offsize / defect.
+    type: string
+    $linkedData:
+      term: offsizeDefect
+      '@id': https://schema.org/description
```
OR13, line 24: I don't suppose there is a better term definition than this?

docs/openapi/components/schemas/common/AgGradeInspectionDefect.yml
```diff
@@ -0,0 +1,61 @@
...
+  averageDefects:
+    title: Average Defects
+    description: Percent of sample items with defects for the grade standards.
+    type: number
+    $linkedData:
+      term: averageDefects
+      '@id': https://schema.org/Number
```
OR13, line 31: I think you are confusing id and type.

docs/openapi/components/schemas/common/AgGradeInspectionLot.yml
```diff
@@ -0,0 +1,255 @@
...
+    $linkedData:
+      term: agProduct
+      '@id': https://w3id.org/traceability#AgProduct
+  lotIdentifier:
+    title: Lot Identifier
+    description: Use to further add to identification of the lot inspected, or to distinguish between lots / sublots for certification purposes (e.g. "Loaded", "Main Lot", "U.S. Fancy", "Film Wrapped").
+    type: string
+    $linkedData:
+      term: lotIdentifier
+      '@id': https://schema.org/identifier
```
OR13, line 31: GS1 maybe?

docs/openapi/components/schemas/common/AgGradeInspectionResult.yml
```diff
@@ -0,0 +1,45 @@
...
+      - type: string
+        const:
+          - AgGradeInspectionResult
+  gradeInspected:
+    title: Grade Inspected
+    description: Grade the lot was inspected for, even if the requirements were not met. When there is no U.S. grade standard for a product, and a quality and condition inspection is performed, enter "No established U.S. Grade".
+    type: string
+    $linkedData:
+      term: gradeInspected
+      '@id': https://w3id.org/traceability#gradeInspected
```
OR13, line 24: We don't currently define this term. https://w3id.org/traceability#gradeInspected

docs/openapi/components/schemas/common/AgGradeInspectionSampleProperty.yml
```diff
@@ -0,0 +1,38 @@
...
+    type: string
+    $linkedData:
+      term: propertValue
+      '@id': https://schema.org/description
+additionalProperties: false
+example: |-
+  {
+    "type": "AgGradeInspectionSampleProperty",
+    "propertyName": "Temp °F",
+    "propertyValue": "53"
```
OR13, line 37: do you want this to be a number or string?

docs/openapi/components/schemas/common/AgParcelDelivery.yml
```diff
@@ -216,7 +216,71 @@ example: |-
...
+          "AgPackage"
+        ],
+        "packageName": "Avocados, Bulk",
+        "grade": "AA",
+        "responsibleParty": {
+          "type": "Entity",
+          "name": "Example Responsible Party Organization",
+          "email": "Chadrick_Gibson@example.com",
+          "phoneNumber": "+1-555-820-1520",
+          "entityType": "Organization"
```
OR13, line 231: This is doing to result in a lot of undefined terms... which will make your RDF not crawl-able.

docs/openapi/components/schemas/common/AgParcelDelivery.yml
```diff
@@ -216,7 +216,71 @@ example: |-
...
+        },
+        "voicePickCode": "4642",
+        "date": "2021-03-14",
+        "labelImageUrl": "https://img.example.org/640/480/",
```
OR13, line 235: why not `label: { image, hash }` ?

docs/openapi/components/schemas/common/AgParcelDelivery.yml
```diff
@@ -216,7 +216,71 @@ example: |-
...
+                "type": [
+                  "QuantitativeValue"
+                ],
+                "unitCode": "hg/ha",
+                "value": "6960"
+              },
+              "sku": "81055399441"
+            },
+            "scientificName": "Persea americana",
+            "labelImageUrl": "https://img.example.org/033383401508/640/480/",
```
OR13, line 275: again with the camel casing over object / predicate definitions... this will make queries awkward.

OR13 (approved):

docs/openapi/components/schemas/common/AgGradeInspectionDefect.yml
rhofvendahl: That makes sense, though I have some reservations.

Do you think I should break from the convention established by AgActivity, AgProduct etc. and call these GradeInspectionDefect, or maybe AgricultureGradeInspectionDefect? Though, I'm concerned the first doesn't indicate that these are specific to agriculture, and the second seems pretty long (especially for something like AgricultureGradeInspectionSampleProperty).

docs/openapi/components/schemas/common/AgGradeInspectionDefect.yml
rhofvendahl: Each AgGradeInspectionDefect represents some type of defect that was observed on a lot during a grade inspection, and offsizeDefect property is a description of that type of defect. So, I think it fits well enough (or at least I don't see a better option on schema.org)

docs/openapi/components/schemas/common/AgGradeInspectionDefect.yml
rhofvendahl: Good to know. So, @id should never refer to a schema.org type? I chose Number because I couldn't find a better property on schema.org (like "percentage"). Do you think schema.org/value would work?

docs/openapi/components/schemas/common/AgGradeInspectionDefect.yml
BenjaminMoe: I think the correct syntax is:

```
        const:
          - AgGradeInspectionDefect
```

The way to check is change type from an array to a string and see if you get an error.

docs/openapi/components/schemas/common/AgGradeInspectionLot.yml
rhofvendahl: Thanks, that seems like a better place to look!

The closest I'm seeing is `gs1.org/voc/hasBatchLotNumber`. I'm a little confused, though, because while Lot Identifier seems like it'd be a good fit, I'm only seeing the gs1 property `hasBatchLotNumber`, which as far as I can tell describes a relationship between `Product` and `string` rather than a class?

Could you provide some guidance? Otherwise I'll just go with `gs1.org/voc/hasBatchLotNumber`, as there seems to be some precedence for using properties (e.g. `gs1.org/voc/variantDescription` on `Package.yml`)

docs/openapi/components/schemas/common/AgGradeInspectionResult.yml
```diff
@@ -0,0 +1,45 @@
...
+    type: string
+    $linkedData:
+      term: details
+      '@id': https://schema.org/description
+additionalProperties: false
+example: |-
+  {
+    "type": "AgGradeInspectionResult",
+    "gradeInspected": "U.S. No. 1",
+    "requirementsMet": true
```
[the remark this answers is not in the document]
rhofvendahl: I'm not sure I understand the difficulty verifying when requirementsMet are false, could you elaborate?

To clarify, requirementsMet is basically whether or not the inspection passed.

docs/openapi/components/schemas/common/AgGradeInspectionSampleProperty.yml
rhofvendahl: This is a bit kludgey but I did intend propertyValue to be a string, even if that string sometimes contains only a number. My thinking was that sample property values really just need to be discernible by humans, and that setting it up to contain either a number or string as appropriate for the unit would result in a convoluted form for inspectors filling it out.

I'm familiar with the use of MeasuredValue & MeasuredProperty, but I don't think those fit super well and IMO would bloat the form.

docs/openapi/components/schemas/common/AgParcelDelivery.yml
rhofvendahl: I don't understand, could you elaborate?

The only change I made was `$ref: ./Product.yml` to `$ref: ./AgPackage.yml`, which given the context I assumed was just fixing a mistake.

docs/openapi/components/schemas/common/AgParcelDelivery.yml
rhofvendahl: Good point, though this seems out of scope for this PR

docs/openapi/components/schemas/common/AgParcelDelivery.yml
rhofvendahl: Again I agree, though I think out of scope

rhofvendahl:
BTW I am investigating the failed check. Looks like there's some issue with npm, which might have just been fixed with https://github.com/w3c-ccg/traceability-vocab/pull/431.

docs/openapi/components/schemas/common/AgGradeInspectionSampleProperty.yml
OR13: https://v.jsld.org/Vjxdn8aYZeKGBe7fe798UpNutDH4JXy9NXbxeGXWNthtNLd4mKZz5SH3mjQsZA5hfe1HotykArBF16gDcCWFvhqXCaAgxXrVEACXRzwWEAY3tZghdvJKmwS4pXomkjU9hDhQsZbtqxudcW6hdNbweTVJaDpq722ykkuYFQgTgNav3R8SCxC7MN3WmwL4yvDNqxA5e1qZmphkzDkdyNP4cjKgDGLry9yRrAb8WdB9PuZtc76WbugvjeM8LkVeqU

```
_:c14n0 <https://w3id.org/traceability/#undefinedTermentityType> "Organization" .
```

docs/openapi/components/schemas/common/AgParcelDelivery.yml
OR13: pretty much you should not be increasing the number of undefined terms in your PR.

docs/openapi/components/schemas/common/AgGradeInspectionDefect.yml
BenjaminMoe: Whenever I see `Ag` my first reaction is "Silver"? Something like `AgricultureGradeInspectionDefect` would be a huge improvement.

docs/openapi/components/schemas/common/AgGradeInspectionResult.yml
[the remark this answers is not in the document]
TallTed: Failed inspections definitely happen, and I think it's even more important that credentials with that payload be verifiable than those for inspections that pass.

docs/openapi/components/schemas/common/AgGradeInspectionResult.yml
rhofvendahl: I think I'm less confused now on the role of @id and the issue with including undefined terms, but I still can't find anything in schema.org (or gs1.org) appropriate for a food grade string.

Would you accept leaving this as is, for lack of a better option?

docs/openapi/components/schemas/common/AgParcelDelivery.yml
rhofvendahl: I'm seeing similar usage of examples with type, name, email, phoneNumber, entityType throughout the vocab. For instance, in BillOfLading (line 176):
```
    "carrier": {
      "type": "Entity",
      "name": "Carrior Goods And More",
      "email": "Adaline29@example.com",
      "phoneNumber": "+1 555-234-9983",
      "entityType": "Organization"
    },
```

I think I understand the issue you're raising but that seems out of scope for this PR. For instance, the block of code above is included as part of an AgPackage example property that was lifted directly from AgPackage, and I don't want to change AgPackage in this PR. I think it would make more sense to create a separate issue to update these entity examples across the board.

So would it work to leave this as-is and create a separate issue?

docs/openapi/components/schemas/common/AgGradeInspectionDefect.yml
nissimsan: @rhofvendahl, `https://qudt.org/vocab/unit/PERCENT` might be a better option for expressing percentage. 

Even better would be something expressing not just the data type, but this particular term. (Related: https://github.com/w3c-ccg/traceability-vocab/issues/230). I would encourage you taking a good look at GS1 EPCIS if there's a suitable definition you can point at.

BenjaminMoe (approved):

docs/openapi/components/schemas/common/AgGradeInspectionDefect.yml
rhofvendahl: Thanks for the tip!

After poking through https://www.gs1.org/voc/ I still haven't find anything relating to percent of items with defects / that are unsuitable for some purpose. qudt.org/vocab/unit/PERCENT definitely seems more appropriate though.

docs/openapi/components/schemas/common/AgParcelDelivery.yml
mkhraisha: For some reason entityType is not being picked up on the Entity objects, I'm not sure why that is.

mkhraisha: That seems like its own issue though, not a blocker on this PR

mkhraisha: Ignore my comments i see you have #440 to fix this

nissimsan (approved):

PR #438, opened 2022-06-14
Adding missing Master BOL type on SeaCargomanifest

nissimsan:
This should fix a bunch of these undefined terms:

image

### DIFF
--- a/docs/openapi/components/schemas/common/SeaCargoManifest.yml
+++ b/docs/openapi/components/schemas/common/SeaCargoManifest.yml
@@ -13,7 +13,8 @@ properties:
           enum:
             - SeaCargoManifest
       - type: string
-        const: SeaCargoManifest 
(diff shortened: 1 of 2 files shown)


BenjaminMoe (approved):

docs/openapi/components/schemas/common/SeaCargoManifest.yml
```diff
@@ -13,7 +13,8 @@ properties:
           enum:
             - SeaCargoManifest
       - type: string
-        const: SeaCargoManifest 
+        const: 
+          - SeaCargoManifest
```
OR13, line 17: is this right?

nissimsan: I don't think it matters. This is what BL and CI are structured.

Issue #272, opened 2022-01-25
use the LEI ontology


VladimirAlexiev:
A LEI ontology has been published maybe 6m ago by @rivettp, @bryon, @dallemang: https://www.gleif.org/ontology/v1.0/L1/index-en.html.
Please use it in the $linkedData mapping of 6 schemas in https://github.com/w3c-ccg/traceability-vocab/tree/main/docs/openapi/components/schemas/common: `LEI*` and `LegalEntityIdentifier*`, as much as reasonable.

Here's a representation of some GLEI L1 data of Fannie Mae:
![image](https://user-images.githubusercontent.com/536250/150996203-9c2e513c-ba9e-4e27-a604-595c270814a0.png)

Specific notes:
- if possible, use GLEI prefixes as shown
- hasValidationSources and hasRegistrationStatus are currently not mapped ok

Maybe related to #232, though I don't know what "LEI prefixing" means. I'll write there.

OR13:
I think we have not defined anything regarding LEI correctly. And we should, but I am not sure we have the attendance or expertise needed to fix this.

BenjaminMoe:
@VladimirAlexiev I made a draft pull request to start using LEI ontology with our Entity schema. Can you provide suggestions for which terms could use `$linkedData` mapping from LEI?

VladimirAlexiev:
Hi @BenjaminMoe! You've mapped to class `gleif-L1:LegalEntity`: that's a start but it's not enough.
- Can you make a list of all classes and props of these 6 schemas? As a comment below. Then I'll try to map them to LEI terms.
- There are also some semantic decisions to make, eg
  - should we use `gleif-L1:legalName` or `schema:name`?
  - if the former, should we declare `gleif-L1:legalName` a subproperty of `schema:name`?

BenjaminMoe:
> Can you make a list of all classes and props of these 6 schemas? As a comment below. Then I'll try to map them to LEI terms.

I'm guessing you mean the six schemas with the `LEI` prefix. Here they are, with their properties listed as bullets. 

1. [LEIaddress.yml](https://github.com/w3c-ccg/traceability-vocab/blob/main/docs/openapi/components/schemas/common/LEIaddress.yml)
    - language
    - addressLines
    - addressNumber
    - addressNumberWithinBuilding
    - mailRouting
    - city
    - region
    - country
    - postalCode
[...]

PR #439, opened 2022-06-14
feat: entity type def

nissimsan:
How about this as an interim term def for entityType? 

I'm sure LEI is the long term answer. But while we figure out https://github.com/w3c-ccg/traceability-vocab/issues/272 it'd be good to remove many, many undefinedTerms. 

PS: Breaks most VC proofs which will have to be re-issued.


nissimsan:
Closing - https://github.com/w3c-ccg/traceability-vocab/pull/440 is better

PR #440, opened 2022-06-14
Fix missing term term in entity

BenjaminMoe:
Assigns a term to `entityType` under the `Entity` schema to avoid undefined terms.


nissimsan (approved):

docs/openapi/components/schemas/common/Entity.yml
```diff
@@ -20,6 +20,9 @@ properties:
     enum:
       - Person
       - Organization
+    $linkedData:
+      term: entityType
+      '@id': https://schema.org/additionalType
```
OR13, line 25: fyi, this is why this does not fix types when you look at RDF view... because the entityType is not type, RDF does not pick up term definitions for it.

OR13 (approved):

mkhraisha (approved):

PR #450, opened 2022-06-20
Fixing a bunch of `undefinedTerm` bugs

nissimsan:
Targeting a bunch of these: https://w3c-ccg.github.io/traceability-vocab/#certificates-with-undefined-terms. Most were banal bugs, and we've just been missing the tooling to catch them earlier.

PR #453, opened 2022-06-21
Weight and Product term fixes

nissimsan:


nissimsan:
The signatures will conflict with https://github.com/w3c-ccg/traceability-vocab/pull/450

docs/openapi/components/schemas/common/TradeLineItem.yml
```diff
@@ -70,17 +70,17 @@ properties:
     description: A measure of the net weight (mass).
     $ref: ./QuantitativeValue.yml
     $linkedData:
-      term: netWeight
+      term: tradeLineItemNetWeight
```
mkhraisha, line 73: is there a reason this and other similar terms like `tradeLineItemGrossWeight` are prefixed by `tradeLineItem`?

Why isn't this just `netWeight` or `grossWeight`?

OR13: This should be fixed... prefxing is against the JSON-LD conventions we have.

OR13 (changes requested):
Address term prefixing, by following the convention

docs/openapi/components/schemas/common/TradeLineItem.yml
nissimsan: Agreed, @mkhraisha , let's stick to the no-prefix convention.

nissimsan:
Agreement to merging once resolved

nissimsan:
Naming convention done, conflict fixed. Merging as agreed.

PR #457, opened 2022-06-27
Remove Mill Test Report RDF Type

BenjaminMoe:
Mill test report is only used in one certificate. As it's not a "common" data type it should only be defined in the credential that uses it. This PR merges Mill test Report into the Mill Test Report Certificate.


OR13 (approved):

mkhraisha (changes requested):
I fundamentally disagree with this PR.
We have a couple of use cases that use the MillTestReport object as a raw schema without having to use the certificate object of it, is this PR change integral to your workflow or is this intended as cleanup?

BenjaminMoe:
@mkhraisha This is a PR that is a proof of concept around data modeling. Right now the schema's in the `common` folder are building blocks for credentials. In the case of `Mill Test Report` and `Mill Test Report Certificate`, this is an example where a building block schema is only used in one credential. And this PR provides an example of how these could be merged.

OR13:
You should add your use cases to the document. The current document only
has a single credential that uses the MTR RDF class.

vivienfan:
I went through it and it won't actually affect us, @mkhraisha 
on the other hand, i think this should extend from `InspectionReport`

mkhraisha (approved):

PR #471, opened 2022-07-12
Schema for ecommerce de minimis exemption

nissimsan:


docs/openapi/components/schemas/credentials/DeMinimisExemptionCertificate.yml
```diff
@@ -0,0 +1,136 @@
...
+    type: object
+additionalProperties: false
+required: []
+example: |-
+  {
+    "@context": [
+      "https://www.w3.org/2018/credentials/v1",
+      "https://w3id.org/traceability/v1"
+    ],
+    "id": "did:key:z6MkrPPZ84figsK3ksWXntFKJGANiJWJYwcusLc9tEKEmgsF",
```
OR13, line 44: stop doing this :)

OR13: use `urn:uuid:1231231` etc...

docs/openapi/components/schemas/credentials/DeMinimisExemptionCertificate.yml
```diff
@@ -0,0 +1,136 @@
...
+        "address": {
+          "type": "PostalAddress",
+          "streetAddress": "19 Giai Phong St., 395 Lane, Phuong Liet Ward",
+          "addressLocality": "Thanh Xuan Dist.",
+          "addressCountry": "VN"
+        },
+        "entityType": "Organization"
+      },
+      "buyer": {
+        "type": "Entity",
```
OR13, line 92: No more using Entity, use Organization.

docs/openapi/components/schemas/credentials/DeMinimisExemptionCertificate.yml
```diff
@@ -0,0 +1,136 @@
...
+        "name": "Phongh Bicycles",
+        "address": {
+          "type": "PostalAddress",
```
OR13, line 95: use Place

OR13 (changes requested):
See comments

docs/openapi/components/schemas/credentials/DeMinimisExemptionCertificate.yml
BenjaminMoe: This is the `issuer.id`, the document needs its own id

OR13: "This" is the credential `id`... it should not be a DID, that is confusing.

TallTed: > "This" is the credential id... it should not be a DID, that is confusing.

Well... It SHOULD (dare I say, MUST?) be a URI, among which DIDs are found, so it *MAY* be a DID, and making it so *MAY* help decrease confusion in the wild though it MAY increase confusion during early adoption and other learning phases.

nissimsan: Happy to change it, if nothing else to make you happy @OR13. :)

OR13: Thank you.

docs/openapi/components/schemas/common/DeMinimisExemption.yml
```diff
@@ -0,0 +1,216 @@
+$linkedData:
+  term: DeMinimisExemption
+  '@id': https://w3id.org/traceability#DeMinimisExemption
+title: De Minimis Exemption
+description: Ecommerce import declaration under Tarif Act Section 321 amendment, providing for an administrative de minimis exemption from duty and taxes for shipments of merchandise (https://www.govinfo.gov/content/pkg/FR-2019-07-23/pdf/2019-15625.pdf).
+type: object
+properties:
+  type:
+    oneOf:
+      - type: array
```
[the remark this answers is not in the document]
nissimsan: I squashed this file into the Certificate schema, so this is gone now.

docs/openapi/components/schemas/credentials/DeMinimisExemptionCertificate.yml
nissimsan: I've changed some of these attributes (`shipmentInitiator` and `seller`), but `buyer` and `finalDeliverTo` will often be a person for De Minimis shipments. Remember, this schema is intended for duties exemption when the import is worth less than $800, specifically for orders made on foreign ecommerce platforms. I propose that we keep these as `Entity` - pls let me know if you still have objections.

docs/openapi/components/schemas/credentials/DeMinimisExemptionCertificate.yml
nissimsan: Yes! Old habits... :)

OR13 (approved):

Issue #192, opened 2021-08-03
PROPOSAL: Define a tagging structure for grouping schema objects


OR13:
- [x] Move industry specific credentials to a folder
- [x] Add OAS tags to each credential

mprorock:
lower priority, but we would like the ability to tag credentials by use case, common usage, etc - closely related to #193 and both could be potentially resolved with a common approach

mprorock:
#220 appears to handle some of this nicely

OR13:
There is a proposal embedded in https://github.com/w3c-ccg/traceability-vocab/pull/220

nissimsan:
I love this!

OR13:
We support this now, via Tags, however we should probably leave this open until we fix the folder structure issues, and are using tags properly.

BenjaminMoe:
We've started with folders. Need to make use of tags in OAS.

BenjaminMoe:
Move to new issues and close.

BenjaminMoe:
I dont think we gain too much advantage from separating credentials by industry folder. I think our folder structure should be `common`, `credentials`, `presentations`. And then we can use OAS tags to sort and search within those folders.

OR13:
I suggest closing, as done.

PR #474, opened 2022-07-13
Add tags to credential schemas

BenjaminMoe:
Partially addresses part of https://github.com/w3c-ccg/traceability-vocab/issues/192.
This is PR add tags to credentials. How they will be surfaced in respec and the openapi will be addressed in a following PR.

### DIFF
--- a/docs/openapi/components/schemas/credentials/CertificateOfOrigin.yml
+++ b/docs/openapi/components/schemas/credentials/CertificateOfOrigin.yml
@@ -2,6 +2,8 @@ $linkedData:
   term: CertificateOfOrigin
   '@id': https://w3id.org/traceability#CertificateOfOrigin
 title: Certificate Of Origin
+tags:
+  - eCommerce
 description: >- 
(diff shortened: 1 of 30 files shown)


nissimsan (approved):

docs/openapi/components/schemas/credentials/CertificateOfOrigin.yml
```diff
@@ -2,6 +2,8 @@ $linkedData:
   term: CertificateOfOrigin
   '@id': https://w3id.org/traceability#CertificateOfOrigin
 title: Certificate Of Origin
+tags:
+  - eCommerce
```
mkhraisha, line 6: This is used in OG, steel, Ecommerce, and AG how do we want to handle tagging for a schema that is used in all?
I'm proposing we add all tags ( obviously out of scope for this PR)

mkhraisha (approved):

docs/openapi/components/schemas/credentials/CertificateOfOrigin.yml
BenjaminMoe: Multiple tags can be added with

```
tags:
  - eCommerce
  - Steel
  - OG
  - AG
```

PR #479, opened 2022-07-15
feat: create schemas & credentials for oft ctes & kdes

rhofvendahl:
This PR creates several schemas & corresponding certificates representing the Critical Tracking Events (CTEs) & Key Data Elements (KDEs) which ag supply chain organizations will be required to report once the upcoming FSMA Traceability Rule takes effect.

Outline of the FSMA Proposed Rule for Food Traceability: https://www.fda.gov/food/food-safety-modernization-act-fsma/fsma-proposed-rule-food-traceability

Details about KDEs required for reporting on the various CTEs: https://www.fda.gov/food/food-safety-modernization-act-fsma/which-key-data-elements-would-apply-me

Mock data for reporting on an example product: https://www.fda.gov/media/145838/download

### DIFF
--- a/docs/openapi/components/schemas/common/FSMAAbstractKDE.yml
+++ b/docs/openapi/components/schemas/common/FSMAAbstractKDE.yml
@@ -1,0 +1,41 @@
+$linkedData:
+  term: FSMAAbstractKDE
+  '@id': https://w3id.org/traceability#FSMAAbstractKDE
+title: FSMA Abstract KDE
+description: >-
+  "An FSMA abstract KDE (Key Data Element), a key/value pair representing a KDE attached to some
+  CTE (Critical Tracking Event). This schema is designed to meet reporting requirements for the FDA
+  Traceability Rule as specified here:
+  https://www.fda.gov/food/food-safety-modernization-act-fsma/fsma-proposed-rule-food-traceability"
+type: object
+properties:
+  type:
+    oneOf:
+      - type: array
+        items:
+          type: string
+          enum:
+            - FSMAAbstractKDE
+      - type: string
+        const: FSMAAbstractKDE
+  name:
+    title: Name
+    description: The name of the KDE.
+    type: string
+    $linkedData:
+      term: name
+      '@id': https://schema.org/propertyID
+  value:
+    title: Value
+    description: The value of the KDE.
+    type: string
+    $linkedData:
+      term: value
+      '@id': https://schema.org/value
+additionalProperties: false
+example: |-
+  {
+    "type": "FSMAAbstractKDE",
+    "name": "Reference record type and number",
+    "value": "BOL 1324"
+  }
--- a/docs/openapi/components/schemas/common/FSMACreatingCTE.yml
+++ b/docs/openapi/components/schemas/common/FSMACreatingCTE.yml
@@ -1,0 +1,125 @@
+$linkedData:
+  term: FSMACreatingCTE
+  '@id': https://w3id.org/traceability#FSMACreatingCTE
+title: FSMA Creating CTE
+description: >-
+  "Information regarding an FSMA Creating CTE (Critical Tracking Event), designed to meet reporting
+  requirements for the FDA Traceability Rule as specified here:
+  https://www.fda.gov/food/food-safety-modernization-act-fsma/fsma-proposed-rule-food-traceability"
+type: object
+properties:
+  type:
+    oneOf:
+      - type: array
+        items:
+          type: string
+          enum:
+            - FSMACreatingCTE
+      - type: string
+        const: FSMACreatingCTE
+  foodCreated:
+    title: Food Created
+    description: The traceability food that was created.
+    $ref: ./FSMAProduct.yml
+    $linkedData:
+      term: foodCreated
+      '@id': https://w3id.org/traceability#FSMAProduct
+  locationCreated:
+    title: Location Created
+    description: Where the product was created.
+    $ref: ./Place.yml
+    $linkedData:
+      term: locationCreated
+      '@id': https://schema.org/Place
+  dateCompleted:
+    title: Date Completed
+    description: The date creation was completed in 24hr ISO 8601 format (e.g. 1970-01-01T00:00Z for UTC, or 1970:01:01T00:00-5:00 for New York on standard time).
+    type: string
+    $linkedData:
+      term: dateCompleted
+      '@id': https://schema.org/DateTime
+  additionalData:
+    title: Additional Data
+    description: Additional Key Data Elements (KDEs).
+    type: array
+    items:
+      $ref: ./FSMAAbstractKDE.yml
+    $linkedData:
+      term: additionalData
+      '@id': https://w3id.org/traceability#FSMAAbstractKDE
+additionalProperties: false
+example: |-
+  {
+    "type": "FSMACreatingCTE",
+    "foodCreated": {
+      "type": "FSMAProduct",
+      "traceabilityLot": {
+        "type": "FSMATraceabilityLot",
+        "lotCode": "2b1ed51f-f1da-49f6-8130-ac81d1e5e61d",
+        "lotCodeAssignmentMethod": "Lot codes are generated as a random UUID.",
+        "lotType": "created"
+      },
+      "quantityUnitMeasure": {
+        "type": [
+          "QuantitativeValue"
+        ],
+        "unitCode": "Crates",
+        "value": "50"
+      },
+      "additionalData": [
+        {
+          "type": "FSMAAbstractKDE",
+          "name": "Category code",
+          "value": "659351253"
+        },
+        {
+          "type": "FSMAAbstractKDE",
+          "name": "Category name",
+          "value": "Nut butter"
+        },
+        {
+          "type": "FSMAAbstractKDE",
+          "name": "Brand name",
+          "value": "Delicious Dips"
+        },
+        {
+          "type": "FSMAAbstractKDE",
+          "name": "Commodity",
+          "value": "Peanut Butter"
+        },
+        {
+          "type": "FSMAAbstractKDE",
+          "name": "Product name",
+          "value": "DD PB"
+        },
+        {
+          "type": "FSMAAbstractKDE",
+          "name": "Packaging size",
+          "value": "24 LB"
+        },
+        {
+          "type": "FSMAAbstractKDE",
+          "name": "Packaging style",
+          "value": "Crates"
+        }
+      ]
+    },
+    "locationCreated": {
+      "type": [
+        "Place"
+      ],
+      "address": {
+        "type": [
+          "PostalAddress"
+        ],
+        "organizationName": "Delicious Dips",
+        "streetAddress": "755 Stephanie Gardens",
+        "addressLocality": "Rasmussenborough",
+        "addressRegion": "KS",
+        "postalCode": "90819",
+        "addressCountry": "United States"
+      },
+      "locationName": "Delicious Dips Production Center"
+    },
+    "dateCompleted": "2022-02-01T00:00Z"
+  }
--- a/docs/openapi/components/schemas/common/FSMACreatingCTE.yml
+++ b/docs/openapi/components/schemas/common/FSMACreatingCTE.yml
@@ -30,14 +30,14 @@ properties:
     $ref: ./Place.yml
     $linkedData:
       term: location
-      '@id': https://schema.org/Place
+      '@id': https://schema.org/location
   dateCompleted:
     title: Date Completed
     description: The date creation was completed in 24hr ISO 8601 format (e.g. 1970-01-01T00:00Z for UTC, or 1970:01:01T00:00-5:00 for New York on standard time).
     type: string
     $linkedData:
       term: dateCompleted
-      '@id': https://schema.org/DateTime
+      '@id': https://schema.org/endDate
   additionalData:
     title: Additional Data
     description: Additional Key Data Elements (KDEs).
--- a/docs/openapi/components/schemas/common/FSMACreatingCTE.yml
+++ b/docs/openapi/components/schemas/common/FSMACreatingCTE.yml
@@ -59,13 +59,8 @@ example: |-
         "lotCodeAssignmentMethod": "Lot codes are generated as a random UUID.",
         "lotType": "created"
       },
-      "quantityUnitMeasure": {
-        "type": [
-          "QuantitativeValue"
-        ],
-        "unitCode": "Crates",
-        "value": "50"
-      },
+      "quantity": 50,
+      "unit": "Crates",
       "additionalData": [
         {
           "type": "FSMAAbstractKDE",
--- a/docs/openapi/components/schemas/common/FSMACreatingCTE.yml
+++ b/docs/openapi/components/schemas/common/FSMACreatingCTE.yml
@@ -17,19 +17,19 @@ properties:
             - FSMACreatingCTE
       - type: string
         const: FSMACreatingCTE
-  foodCreated:
-    title: Food Created
+  food:
+    title: Food
     description: The traceability food that was created.
     $ref: ./FSMAProduct.yml
     $linkedData:
-      term: foodCreated
+      term: food
       '@id': https://w3id.org/traceability#FSMAProduct
-  locationCreated:
-    title: Location Created
+  location:
+    title: Location
     description: Where the product was created.
     $ref: ./Place.yml
     $linkedData:
-      term: locationCreated
+      term: location
       '@id': https://schema.org/Place
   dateCompleted:
     title: Date Completed
@@ -51,7 +51,7 @@ additionalProperties: false
 example: |-
   {
     "type": "FSMACreatingCTE",
-    "foodCreated": {
+    "food": {
       "type": "FSMAProduct",
       "traceabilityLot": {
         "type": "FSMATraceabilityLot",
@@ -104,7 +104,7 @@ example: |-
         }
       ]
     },
-    "locationCreated": {
+    "location": {
       "type": [
         "Place"
       ],
--- a/docs/openapi/components/schemas/common/FSMAFirstReceiverData.yml
+++ b/docs/openapi/components/schemas/common/FSMAFirstReceiverData.yml
@@ -105,8 +105,8 @@ example: |-
         "type": "Entity",
         "entityType" : "Person",
         "name" : "John Davis",
-        "email": "jdavis@johnsproduce.com",
-        "phoneNumber": "129.697.0673"
+        "email": "produce@example.com",
+        "phoneNumber": "129-697-0673"
       },
       "lotType": "originated"
     },
--- a/docs/openapi/components/schemas/common/FSMAFirstReceiverData.yml
+++ b/docs/openapi/components/schemas/common/FSMAFirstReceiverData.yml
@@ -1,0 +1,133 @@
+$linkedData:
+  term: FSMAFirstReceiverCTE
+  '@id': https://w3id.org/traceability#FSMAFirstReceiverCTE
+title: FSMA First Receiver Data
+description: >-
+  "Information regarding FSMA First Receiver data, a collection of KDEs (Key Data Elements) for
+  which a product's first receiver is responsible. This schema is designed to meet reporting
+  requirements for the FDA Traceability Rule as specified here:
+  https://www.fda.gov/food/food-safety-modernization-act-fsma/fsma-proposed-rule-food-traceability"
+type: object
+properties:
+  type:
+    oneOf:
+      - type: array
+        items:
+          type: string
+          enum:
+            - FSMAFirstReceiverCTE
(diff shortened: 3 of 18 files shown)


mprorock (approved):
Awesome work

OR13:
woah.

docs/openapi/components/schemas/common/FSMAAbstractKDE.yml
```diff
@@ -0,0 +1,41 @@
...
+            - FSMAAbstractKDE
+      - type: string
+        const: FSMAAbstractKDE
+  name:
+    title: Name
+    description: The name of the KDE.
+    type: string
+    $linkedData:
+      term: name
+      '@id': https://schema.org/propertyID
```
OR13, line 27:
```suggestion
      '@id': https://schema.org/name
```

docs/openapi/components/schemas/common/FSMACreatingCTE.yml
```diff
@@ -0,0 +1,125 @@
...
+      - type: string
+        const: FSMACreatingCTE
+  foodCreated:
+    title: Food Created
+    description: The traceability food that was created.
+    $ref: ./FSMAProduct.yml
+    $linkedData:
+      term: foodCreated
+      '@id': https://w3id.org/traceability#FSMAProduct
+  locationCreated:
```
OR13, line 27: prefer `location`.... you are repeating `Created` a lot....

docs/openapi/components/schemas/common/FSMACreatingCTE.yml
```diff
@@ -0,0 +1,125 @@
...
+    $linkedData:
+      term: locationCreated
+      '@id': https://schema.org/Place
+  dateCompleted:
+    title: Date Completed
+    description: The date creation was completed in 24hr ISO 8601 format (e.g. 1970-01-01T00:00Z for UTC, or 1970:01:01T00:00-5:00 for New York on standard time).
+    type: string
+    $linkedData:
+      term: dateCompleted
+      '@id': https://schema.org/DateTime
```
OR13, line 40: id type confusion.

OR13: Find a `property` from this list: 

https://schema.org/Date

for example:

https://schema.org/dateCreated
https://schema.org/productionDate

docs/openapi/components/schemas/common/FSMACreatingCTE.yml
```diff
@@ -0,0 +1,125 @@
...
+        "type": "FSMATraceabilityLot",
+        "lotCode": "2b1ed51f-f1da-49f6-8130-ac81d1e5e61d",
+        "lotCodeAssignmentMethod": "Lot codes are generated as a random UUID.",
+        "lotType": "created"
+      },
+      "quantityUnitMeasure": {
+        "type": [
+          "QuantitativeValue"
+        ],
+        "unitCode": "Crates",
```
OR13, line 66: sadly... not a unit code... https://schema.org/unitCode

OR13: I think we have this same problem on several schemas.

docs/openapi/components/schemas/common/FSMAFirstReceiverData.yml
```diff
@@ -0,0 +1,133 @@
...
+    $linkedData:
+      term: originatorLocation
+      '@id': https://schema.org/Place
+  harvestDate:
+    title: Harvest Date
+    description: The date and time product was harvested in 24hr ISO 8601 format (e.g. 1970-01-01T00:00Z for UTC, or 1970:01:01T00:00-5:00 for New York on standard time).
+    type: string
+    $linkedData:
+      term: harvestDate
+      '@id': https://schema.org/DateTime
```
OR13, line 41: id / type confusion

docs/openapi/components/schemas/common/FSMAFirstReceiverData.yml
```diff
@@ -0,0 +1,133 @@
...
+    $linkedData:
+      term: coolingDate
+      '@id': https://schema.org/DateTime
+  packingLocation:
+    title: Packing Location
+    description: Where the food was packed.
+    $ref: ./Place.yml
+    $linkedData:
+      term: packingLocation
+      '@id': https://schema.org/Place
```
OR13, line 62: id / type confusion

docs/openapi/components/schemas/common/FSMAGrowingCTE.yml
```diff
@@ -0,0 +1,84 @@
...
+        const: FSMAGrowingCTE
+  traceabilityLot:
+    title: Traceability Lot
+    description: The lot of food being tracked.
+    $ref: ./FSMATraceabilityLot.yml
+    $linkedData:
+      term: traceabilityLot
+      '@id': https://w3id.org/traceability#FSMATraceabilityLot
+  growingAreaCoordinates:
+    title: Growing Area Coordinates [TODO this could be attached to lot directly removing need for this schema - but should it? Won't be used by other schemas that ref trace lot... I think?]
```
OR13, line 28: move TODO to an issue...

docs/openapi/components/schemas/common/FSMAGrowingCTE.yml
```diff
@@ -0,0 +1,84 @@
...
+    description: Coordinates of the area the product was grown.
+    $ref: ./GeoCoordinates.yml
+    $linkedData:
+      term: growingAreaCoordinates
+      '@id': https://schema.org/GeoCoordinates
```
OR13, line 33: id / type confusion

docs/openapi/components/schemas/common/FSMAGrowingCTE.yml
```diff
@@ -0,0 +1,84 @@
...
+          "postalCode": "67391",
+          "addressCountry": "United States"
+        },
+        "locationName": "John's Tomato Farm #1"
+      },
+      "lotCodeGeneratorPOC": {
+        "type": "Entity",
+        "entityType" : "Person",
+        "name" : "John Davis",
+        "email": "jdavis@johnsproduce.com",
```
OR13, line 72: is this a real email?

docs/openapi/components/schemas/common/FSMAGrowingCTE.yml
```diff
@@ -0,0 +1,84 @@
...
+          "addressLocality": "Port Mark",
+          "addressRegion": "LA",
+          "postalCode": "67391",
+          "addressCountry": "United States"
+        },
+        "locationName": "John's Tomato Farm #1"
+      },
+      "lotCodeGeneratorPOC": {
+        "type": "Entity",
+        "entityType" : "Person",
```
OR13, line 70: You are gonna end up with a bunch of undefined terms... thats going to be data that is not processable or mergeable...

docs/openapi/components/schemas/common/FSMAGrowingCTE.yml
```diff
@@ -0,0 +1,84 @@
...
+        "name" : "John Davis",
+        "email": "jdavis@johnsproduce.com",
+        "phoneNumber": "129.697.0673"
```
OR13, line 73: lol `-` vs `.`

docs/openapi/components/schemas/common/FSMAProduct.yml
```diff
@@ -0,0 +1,127 @@
...
+    $linkedData:
+      term: traceabilityLot
+      '@id': https://w3id.org/traceability#FSMATraceabilityLot
+  quantityUnitMeasure:
+    title: Quantity and unit of measure
+    description: The quantity and unit measure of this product (e.g. 10 Cartons, or 20 Crates).
+    $ref: ./QuantitativeValue.yml
+    $linkedData:
+      term: quantityUnitMeasure
+      '@id': https://schema.org/QuantitativeValue
```
OR13, line 34: id / type confusion.

docs/openapi/components/schemas/common/FSMAProduct.yml
```diff
@@ -0,0 +1,127 @@
...
+      },
+      {
+        "type": "FSMAAbstractKDE",
+        "name": "Variety",
+        "value": "Mighty Sweet Hybrid"
+      },
+      {
+        "type": "FSMAAbstractKDE",
+        "name": "Product name",
+        "value": "N/A"
```
OR13, line 114: why you include an `N/A` here... ?

docs/openapi/components/schemas/common/FSMAProduct.yml
```diff
@@ -0,0 +1,127 @@
...
+      "lotType": "originated"
+    },
+    "quantityUnitMeasure": {
+      "type": [
+        "QuantitativeValue"
+      ],
+      "unitCode": "Boxes",
+      "value": "82"
+    },
+    "additionalData": [
```
OR13, line 85: This is maybe not how I would model this... I would be happy to learn more about this use case though.

docs/openapi/components/schemas/common/FSMAReceivingCTE.yml
```diff
@@ -0,0 +1,194 @@
...
+    $linkedData:
+      term: shipment
+      '@id': https://w3id.org/traceability#FSMAShipment
+  dateReceived:
+    title: Date Received
+    description: The date shipment was received in 24hr ISO 8601 format (e.g. 1970-01-01T00:00Z for UTC, or 1970:01:01T00:00-5:00 for New York on standard time).
+    type: string
+    $linkedData:
+      term: dateReceived
+      '@id': https://schema.org/DateTime
```
OR13, line 33: id / type confusion

docs/openapi/components/schemas/common/FSMAReceivingCTE.yml
```diff
@@ -0,0 +1,194 @@
...
+  additionalData:
+    title: Additional Data
+    description: Additional Key Data Elements (KDEs).
+    type: array
+    items:
+      $ref: ./FSMAAbstractKDE.yml
+    $linkedData:
+      term: additionalData
+      '@id': https://w3id.org/traceability#FSMAAbstractKDE
```
OR13, line 42: id / type confusion

docs/openapi/components/schemas/common/FSMAReceivingCTE.yml
```diff
@@ -0,0 +1,194 @@
...
+              "addressLocality": "Port Mark",
+              "addressRegion": "LA",
+              "postalCode": "67391",
+              "addressCountry": "United States"
+            },
+            "locationName": "John's Tomato Farm #1"
+          },
+          "lotCodeGeneratorPOC": {
+            "type": "Entity",
+            "entityType" : "Person",
```
OR13, line 74: You will have lots of undefined terms

docs/openapi/components/schemas/common/FSMAReceivingCTE.yml
```diff
@@ -0,0 +1,194 @@
...
+            "value": "Bulk"
+          },
+          {
+            "type": "FSMAAbstractKDE",
+            "name": "Packaging style",
+            "value": "Field Bins"
+          }
+        ]
+      },
+      "from": {
```
OR13, line 131: interesting semantics, why not `source` and `destination`?

docs/openapi/components/schemas/common/FSMAShipment.yml
```diff
@@ -0,0 +1,198 @@
...
+    $linkedData:
+      term: from
+      '@id': https://schema.org/Place
+  to:
+    title: Ships To
+    description: The shipment's immediate destination.
+    $ref: ./Place.yml
+    $linkedData:
+      term: to
+      '@id': https://schema.org/Place
```
OR13, line 41: id / type... what you mean to use here is:

The property with `id`: https://schema.org/location
Which supports the `type: https://schema.org/Place

This mistake is all over the place (not just in your PR here, but in many other schemas).

Here is an example of what I mean:

https://v.jsld.org/6DxQAm4Rf3J61R2L8HSynh9Hg741hKe4F4vJU1Z7VbdYnDPyx3WyzjuBcr5UMHpLtbis7Xsx7AjbWgYqaPK4fNh86qdiqvaJ7eN7GTTvRPiYoRR3aYziZ6z89KrYXDMCE1116SiKGXynhVJzvWBHcotgSemPyzaETLWCX5z2UeeMxsuRmgnn7s1qf8xRMaXDqT41EoDEZcps3L61c33oxjtqZZduTDhWD2Sw6xDMEYZATLmrh1rCE65Xwj6Ahpg8cpFmQXUVPygtAPShnZwbKXuzCZPfo9tYGkksEiNbrNLKB6YXaaKEdiDvxSXZaqodyF9nacxkgR58VB2Yn8PcRK2xvHZ2KH8JhmvPrjT8K9yYJ8riev2s9npwsGie2HgY2kxgpuKz6nFS2Ddad2QTxVBKGKR2eZjZXQGpc4wsY7qojJkCukn85Zj49Pw5kyPsLJiiqWnKpgRiW8

OR13: The example is taken from: https://schema.org/Place#eg-0171

OR13:
```
_:c14n0 <https://schema.org/location> _:c14n3 ._:c14n0 <https://schema.org/location> _:c14n3 .
...
_:c14n3 <https://schema.org/address> _:c14n2 .
... 
_:c14n2 <https://schema.org/postalCode> "80209" .
```

OR13: You only get "machine readability" if your JSON-LD /  RDF is correct... that means using `id` and `type` correctly.

OR13 (approved):
This PR is massive.

I am approving it in the interest of seeing SMALLER PRs which refine the various errors which I have commented on, which will remain observable after it is merged.

The main thing to learn about is here:

https://github.com/w3c-ccg/traceability-vocab/pull/479/files#r922547127

After this PR is merged, this chart will show a massive amount of failure:

https://w3c-ccg.github.io/traceability-vocab/#certificates-with-undefined-terms

Screen Shot 2022-07-15 at 5 14 58 PM

We have agreed not to hold up PRs due to undefined terms, but know that you are injecting a huge amount of undefined terms in this PR and in the future, we may resolve to exclude ANY credentials that do not define their terms fully from interoperability tests... this is because the graph processing value of credentials with poorly defined terms is drastically diminished, which is something I believe we are trying to support in the long term.

(We have similar issues in our schemas, but its a bad habit we all need to kick).

docs/openapi/components/schemas/common/FSMAShipment.yml
mprorock: Yes @rhofvendahl this should be corrected

docs/openapi/components/schemas/common/FSMAReceivingCTE.yml
mprorock: This is FDA language unfortunately

docs/openapi/components/schemas/common/FSMAReceivingCTE.yml
mprorock: +1

docs/openapi/components/schemas/common/FSMAProduct.yml
mprorock: A lot of this is because of PMI and FDA - we are not fans but need to support the regs as written

docs/openapi/components/schemas/common/FSMAProduct.yml
mprorock: That is straight from FDA

docs/openapi/components/schemas/common/FSMAGrowingCTE.yml
mprorock: Should use examp.org

docs/openapi/components/schemas/common/FSMACreatingCTE.yml
mprorock: In this case FDA uses Unit code this way - we should say `FSMAUnitCode` or something like that @rhofvendahl

OR13: How big is the coding system? We can define them all here possibly.
Or point to an existing ontology that defines them...

No hits in https://www.gs1.org/voc/

Have you checked https://qudt.org/ ?

nissimsan:
@rhofvendahl the `FSMA` acronym isn't spelled out anywhere. I suggest you spell it out on the various schema `description`s. 
Great work!  :)

docs/openapi/components/schemas/common/FSMACreatingCTE.yml
rhofvendahl: @OR13 the units are a little fuzzy, e.g. "crates", "pallets", "field bins".

@mprorock  correct me if I'm wrong but I think while standardized units for weight or volume might be useful, they're not currently part of the suggested reporting workflow (I'm guessing because users might not have easy access to that info &/or reviewers might not find that info particularly useful).

Looks like `https://schema.org/unitText` provides an alternative for units that don't have a standard unit code, which seems great for our purposes. Instead of referring to a QuantitativeValue I'll just provide `quantity` and `unit` properties where unit has id `unitText`.

docs/openapi/components/schemas/common/FSMAFirstReceiverData.yml
```diff
@@ -0,0 +1,133 @@
...
+            - FSMAFirstReceiverCTE
+      - type: string
+        const: FSMAFirstReceiverCTE
+  traceabilityLot:
+    title: Traceability Lot
+    description: The lot of food being tracked.
+    $ref: ./FSMATraceabilityLot.yml
+    $linkedData:
+      term: traceabilityLot
+      '@id': https://w3id.org/traceability#FSMATraceabilityLot
```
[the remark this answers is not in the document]
rhofvendahl: My understanding is that hasBatchLotNumber is a string property attached to something like a product, whereas this traceabilityLot property is a ref to a bunch of lot data.

I do use hasBatchLotNumber as the id of the lotCode property within FSMATraceabilityLot

docs/openapi/components/schemas/common/FSMAGrowingCTE.yml
rhofvendahl: Yikes, thanks 😬

docs/openapi/components/schemas/common/FSMAGrowingCTE.yml
rhofvendahl: It is not a real email but I'll use produce@example.com to make that clear

docs/openapi/components/schemas/common/FSMAReceivingCTE.yml
rhofvendahl: Not very important but FWIW it's also the most clear language I can think of, or else I'd recommend shifting it 🤷‍♂️

`to` and `from` are properties on the FSMAShipment schema, which has a lot of data common to both `FSMAReceivingCTE` and `FSMAShippingCTE`. In both cases the `to` and `from` refer to the immediate previous destination and immediate subsequent destination of that leg of the journey.

The wording of the KDEs for both shipping & receiving is super clear about _immediate_ previous/subsequent locations, presumably so that the "source" for a product that's been shipped a lot is taken to mean the previous stop rather than the farm it was harvested on or something. So, whatever wording used has to imply this.

Given that `shipment` is used for KDEs shared between `shipping` and `receiving`, to/from seemed more clear than immediatePreviousSource/immediateNextDestination, or immediateSource/immediateDestination.

docs/openapi/components/schemas/common/FSMAShipment.yml
rhofvendahl: @OR13 Thanks for this info! I definitely let laziness get the best of me with id/type in this PR, I'll make that a priority going forward.

rhofvendahl:
@OR13 thanks for bringing attention to the undefined terms issue. Regarding suggested fixes I'll submit changes to the minor stuff and leave bigger changes to subsequent PRs, as you suggest.

@nissimsan Thanks for the tip, I've added that info to certificate descriptions only to keep things succinct :-)

docs/openapi/components/schemas/common/FSMAGrowingCTE.yml
rhofvendahl: @OR13 This example text represents a reference to our current Entity schema. I'm not sure I'm following, what about that schema will result in undefined terms?

docs/openapi/components/schemas/common/FSMAReceivingCTE.yml
rhofvendahl: @OR13 This is another case where I am trying to reference our Entity schema. Is there a better way to do this?

mkhraisha (approved):

PR #480, opened 2022-07-15
Fix Intent to Import

OR13:
rename `IntentToSell` to `IntentToImport`.

fix undefined terms.

add better term definitions.

Note this is the RDF Class, NOT a Verifiable Credential.

We will need to add a Verifiable Credential for this type... that example will have a `proof` and an `issuer`... properties which are not present in this RDF Class.


mprorock (approved):

docs/openapi/components/schemas/common/IntentToImport.yml
```diff
@@ -1,7 +1,7 @@
 $linkedData:
-  term: IntentToSell
-  '@id': https://w3id.org/traceability#IntentToSell
-title: Intent To Sell
+  term: IntentToImport
+  '@id': https://w3id.org/traceability#IntentToImport
+title: Intent To Import
 description: >-
   A simple case where an entity (person or organization) wishes to notify some
   third party that they intend to sell some product, likely into a different
```
TallTed, line 7: This will nearly duplicate the second line, which has `in` where `it` should be. The originator of this PR should be able to correct that.
```suggestion
  third party that they intend to import some product, likely into a different
  geographic region than it originated in.  This is a highly simplified version
```

docs/openapi/components/schemas/common/IntentToImport.yml
```diff
@@ -53,20 +53,20 @@ properties:
     type: string
     $linkedData:
       term: declarationDate
-      '@id': https://schema.org/DateTime
+      '@id': https://schema.org/startDate
   sellByDate:
     title: Sell By Date
     description: Date delcared as a non-binding target to sell product by.
```
TallTed, line 59:
```suggestion
    description: Date declared as a non-binding target to sell product by.
```

BenjaminMoe (approved):

docs/openapi/components/schemas/common/IntentToImport.yml
```diff
@@ -1,10 +1,11 @@
...
-  '@id': https://w3id.org/traceability#IntentToSell
-title: Intent To Sell
+  term: IntentToImport
+  '@id': https://w3id.org/traceability#IntentToImport
+title: Intent To Import
 description: >-
   A simple case where an entity (person or organization) wishes to notify some
-  third party that they intend to sell some product, likely into a different
+  third party that they intend to import some product, likely into a different
+  geographic region than it originated in.  This is a highly simplified version
```
TallTed, line 8:
```suggestion
  geographic region than that of its origin.  This is a highly simplified version
```
Or perhaps, `than where it originated.`

nissimsan (approved):

mkhraisha (approved):

mkhraisha (approved):

PR #490, opened 2022-07-20
fix: description wording

nissimsan:
Includes the wording suggestions in https://github.com/w3c-ccg/traceability-vocab/pull/480 

@TallTed  - FYI and thanks!


nissimsan:
We looked at this already, and it's not particularly substantial, so taking the liberty to just merge this in.

PR #492, opened 2022-07-20
Rename to "Certification of Origin"

nissimsan:
Based on feedback from DHS, changing from "Certificate of Origin" to "Certification of Origin".


mprorock (approved):

nissimsan:
Changing this to Draft. We're confirming if `Certification` is indeed a better term.

mprorock:
> Changing this to Draft. We're confirming if `Certification` is indeed a better term.

we are OK either way - "Certificate of Origin" we encounter more often in Ag

BenjaminMoe (approved):

nissimsan:
> > Changing this to Draft. We're confirming if `Certification` is indeed a better term.
> 
> we are OK either way - "Certificate of Origin" we encounter more often in Ag

Yes, "Certificate" is also the prevalent term in shipping and what I am used to. But we confirmed "Certification" with CBP now so let's go for that.

mkhraisha (approved):

nissimsan:
Let me look into why the file delete is not here. 
Merging offline.

nissimsan:
Fixed. Merging.

PR #503, opened 2022-07-26
feat: create ThirdPartyClaimsCertificate & supporting schemas

rhofvendahl:
This PR creates a certificate for claims made by a third party (as requested by @mprorock ).

### DIFF
--- a/docs/openapi/components/schemas/common/Claim.yml
+++ b/docs/openapi/components/schemas/common/Claim.yml
@@ -1,0 +1,44 @@
+$linkedData:
+  term: Claim
+  '@id': https://w3id.org/traceability#Claim
+title: Claim
+description: A claim's subject and value.
+type: object
+properties:
+  type:
+    oneOf:
+      - type: array
+        items:
+          type: string
+          enum:
+            - Claim
+      - type: string
+        const: Claim
+  claim:
+    title: Claim
+    description: The subject of a claim.
+    type: string
+    $linkedData:
+      term: claim
+      '@id': https://schema.org/name
+  value:
+    title: Value
+    description: The value of a claim.
+    type: string
+    $linkedData:
+      term: value
+      '@id': https://schema.org/value
+  valueObject:
+    title: Value
+    description: The object value of a claim.
+    type: object
+    $linkedData:
+      term: valueObject
+      '@id': https://schema.org/value
+additionalProperties: false
+example: |-
+  {
+    "type": "Claim",
+    "claim": "ImporterOfRecordNumber",
+    "value": "12-3456789"
+  }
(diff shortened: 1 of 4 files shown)


docs/openapi/components/schemas/common/Claim.yml
```diff
@@ -0,0 +1,44 @@
...
+      term: claim
+      '@id': https://schema.org/name
+  value:
+    title: Value
+    description: The value of a claim.
+    type: string
+    $linkedData:
+      term: value
+      '@id': https://schema.org/value
+  valueObject:
```
mkhraisha, line 31: is this the same as Value just as an object not a string?

mkhraisha (approved):

docs/openapi/components/schemas/common/Claim.yml
rhofvendahl: Yeah, Mike said values would be either strings or objects & this seemed like a reasonable approach.

docs/openapi/components/schemas/credentials/ThirdPartyClaimsCertificate.yml
```diff
@@ -0,0 +1,81 @@
+$linkedData:
+  term: ThirdPartyClaimsCertificate
+  '@id': https://w3id.org/traceability#ThirdPartyClaimsCertificate
+title: Third Party Claims Certificate
```
BenjaminMoe, line 4: Add tag?

nissimsan:
@mprorock, we discussed this a little on the call. The intention of this PR confuses me. This looks like generic claims, but isn't that what a VerifiableCredential already is? 
What am I missing?

mprorock (approved):

mprorock:
> @mprorock, we discussed this a little on the call. The intention of this PR confuses me. This looks like generic claims, but isn't that what a VerifiableCredential already is? What am I missing?

It totally is.  This is based on a request fr two different customers to provide a formal schema for description of a link between a system external to trace and a traceable presentation in a repeatable and flexible manner so that it facilitates an easier upgrade path with opaque customer systems

TallTed:
This feels to me to be essentially a new costume for a Verifiable Presentation (VP)... 

Most of the use case appears to be business logic to handle such a VP, which we *could* standardize, but I wonder whether that won't effectively impose limitations that diminish the applicability and utility of VPs for entities that are not currently participating in this standardization process.

(I also wonder why this began as a PR, and not an Issue.)

OR13 (changes requested):
This seems like an attempt to reinvent JSON-LD on JSON-LD.

nissimsan:
Re-inventing https://www.w3.org/TR/vc-data-model/#claims ?

nissimsan:
General pushback on the trace call. Suggestion to bring this up on an issue first.

OR13:
We also discussed how https://www.w3.org/TR/vc-data-model/#evidence might be a useful way to represent this.

rhofvendahl:
@brownoxford Yes, doing that now.

Issue #374, opened 2022-04-12
Evidence, Related Links, Related Documents and Hashlinks what should we use?


mkhraisha:
This issue is meant as a continuation of the thread on the mailing list with the goal of attempting to find a solution that we can standardize around in this vocabulary.

# The Issue:
How do we reference a supply chain credential from another credential?
To illustrate the use case consider a Transportation Event Credential that represents a Transportation Event that a Product underwent, and has an associated Bill Of Lading Credential. The Transportation Event Credential wants to reference the Product  Credential and the Bill of Lading Credential.

There are two parts to this question, how can we reference other credentials and where can we put that reference.

## Ways to reference other credentials:
1.  URI that points to a hosted VC with the Product and Bill Of Lading objects
```
{
    // rest of VC
    "credentialSubject": {
        "type": [
            "TransportationEvent"
        ],
        // rest of eventVC
        "product": "example.com/product/123",
        "BillOfLading": "example.com/billoflading/123"
    }
}
```
2. [Hashlink](https://w3c-ccg.github.io/hashlink/) that points to the hosted VCs
```
{
    // rest of VC
    "credentialSubject": {
        "type": [
            "TransportationEvent"
        ],
        // rest of eventVC
        "product": "http://example.com/product/123?hl=zQmWvQxTqbG2Z9HPJgG57jjwR154cKhbtJenbyYTWkjgF3e",
        "BillOfLading": "http://example.org/billoflading/123?hl=zQmWvQxTqbG2Z9HPJgG57jjwR154cKhbtJenbyYTWkjgF3e"
    }
}
```

3. Directly embed the other credential // this seems pretty awful

## Places to put the reference
1. In the credential directly under a vocabulary defined property such as `product` or `bill of lading`
```
{
    // rest of VC
    "credentialSubject": {
        "type": [
            "TransportationEvent"
        ],
        // rest of eventVC
        "product": "REFERENCE HERE"
        "BillOfLading": "REFERENCE HERE"
    }
}
```
2. In the [Evidence](https://www.w3.org/TR/vc-data-model/#evidence) property as defined by the VC data model
```
{
    // rest of VC
    "credentialSubject": {
           "type": [
            "TransportationEvent"
        ],
        // rest of eventVC
    },
    "evidence": [
    // obviously this will need to be compliant with VC data model evidence property
        {
          "REFERENCE HERE for product"
        },
        {
            "REFERENCE HERE for the event"
        }
    ]
}
```

[...]

PR #504, opened 2022-07-26
Referencing VC

vivienfan:
Related issue: https://github.com/w3c-ccg/traceability-vocab/issues/374

Replacing directly embed credential with a hashlink which points to the hosted VCs

`Hashlink` is prefered over `Evidence` and `Related Link` because:
1. it makes sure the data being referenced is the same
    - in the case of `Events`, `Products` should stay the same (ie, size, amount, manufactuer, etc.)
    - `Hashlink` makes sure the content returned by the link is the same as when it's generated, otherwise the hash would change
    - `Evidence` and `Related Link` can only meet this requirement if `reference`/`target` is `Hashlink`

2. it describes the relationship between `Products` and `Events` in a dry way
    - in the case of `Transform Events`, there are 2 types of `Products` - newly created & consumed/used products
    - `Hashlink` is replacing the directly embed credential in the structure, which already describes the type of `Products`
    - `Evidence` does not describe the type of credential being referenced
    - `Related Link` has a field `LinkRelationship` to meet this requirement, but the solution is more complicated and harder to read

### DIFF
--- a/docs/openapi/components/schemas/common/CommissionEvent.yml
+++ b/docs/openapi/components/schemas/common/CommissionEvent.yml
@@ -40,14 +40,13 @@ properties:
       '@id': https://w3id.org/traceability#Organization
   products:
     title: Products
-    description: >-
-      List of product details as outlined in the Product schema
+    description: The products referenced by the event in hashlinks.
     type: array
     items:
-      $ref: ./Product.yml
+      type: string
     $linkedData:
-      term: product
-      '@id': https://schema.org/Product
+      term: products
+      '@id': https://w3c-ccg.github.io/hashlink/#hl-url-params
 additionalProperties: false
 example: |-
   {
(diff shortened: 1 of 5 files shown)


nissimsan (changes requested):

nissimsan:
@vivienfan, so the intention is because there are multiple events for the same product, you don't want to repeat the product data? That's a good intention. 
Signing a link doesn't seem very valuable, though. I would suggest signing over the product's id instead?

vivienfan:
> @vivienfan, so the intention is because there are multiple events for the same product, you don't want to repeat the product data? That's a good intention. Signing a link doesn't seem very valuable, though. I would suggest signing over the product's id instead?

@nissimsan, the link here would be a `Get` request endpoint which returns the product VC, and the hash would make sure the product is the exact same product when the event VC was issued.

nissimsan:
@vivienfan I see. That's a new pattern, but I won't stand in the way. :) Thanks for elaborating.

nissimsan (approved):

mkhraisha (approved):

PR #507, opened 2022-07-29
Json schema validation errors

nissimsan:
This fixes a ton of the JSON Schema errors. These are the ones which are normally hidden, but emerge on the console when other errors exist. 
There are three main category of bugs fixed: 
- Array `const` instead of `enum`, which has been copy-pasted many times
- Changes to Org and Entity propagating to many places where used  
- "Normal", one-off JSON Schema bugs (vast majority mine)


nissimsan:
This is getting too big - slicing it into smaller chunks...

brownoxford:
@nissimsan can we delete the branch?

PR #514, opened 2022-08-03
Git ignore `results.js`

nissimsan:
results.js is generated from `npm run t`, it causes thrashing on PRs.


brownoxford:
@nissimsan can we delete this branch?

PR #519, opened 2022-08-09
Stale proofs

BenjaminMoe:
These are proofs that `regnerate.js` automatically regenerated. Only runs once, after that they are detected as valid.


nissimsan (approved):

brownoxford:
@BenjaminMoe can this branch be deleted?

PR #520, opened 2022-08-09
Update organizationName to name in PostalAddress

Ralhusban:
Change organizationName to name in the PostalAddress Object. The organizationName is not generic enough to represent all objects that use Postal Address object. They are not necessarely organizations as they can represent any entity. For example, Schema.org uses name inside the Postal Address schema.

### DIFF
--- a/docs/openapi/components/schemas/common/AgricultureInspectionReport.yml
+++ b/docs/openapi/components/schemas/common/AgricultureInspectionReport.yml
@@ -119,7 +119,7 @@ example: |-
         "type": [
           "PostalAddress"
         ],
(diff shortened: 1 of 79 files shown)


OR13 (approved):

mkhraisha (approved):

nissimsan (approved):

nissimsan:
@Ralhusban, reminder to re-proof the certificates to make the checks pass. When you change the term definition, you break the credential signatures where PostalAdr is used.

OR13:
CI must be passing prior to merging

Ralhusban:
This PR was replaced by PR #533

PR #524, opened 2022-08-16
Removing type oneOf (array or string)

nissimsan:
This prioritizes consistency and readability over flexibility by removing optionality of specifying types as a basic strings. 

This is quite abstract to decipher: 
```
  type:
    oneOf:
      - type: array
        items:
          type: string
          enum:
            - MyTypeName
      - type: string
        const: MyTypeName 
```

I've only touched "our" schemas. I encourage y'all to follow the same pattern for consistency and simplification.

### DIFF
--- a/docs/openapi/components/schemas/common/AdditionalProductCodeRegistrationCredential.yml
+++ b/docs/openapi/components/schemas/common/AdditionalProductCodeRegistrationCredential.yml
@@ -7,14 +7,11 @@ description: Ecommerce Additional Product Code Registration Credential
 type: object
 properties:
   type:
-    oneOf:
-      - type: array
-        items:
-          type: string
-          enum:
-            - AdditionalProductCodeRegistrationCredential
-      - type: string
-        const: AdditionalProductCodeRegistrationCredential
+    type: array
+    items:
+      type: string
+      enum:
+        - AdditionalProductCodeRegistrationCredential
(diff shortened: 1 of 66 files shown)


docs/openapi/components/schemas/common/AdditionalProductCodeRegistrationCredential.yml
```diff
@@ -7,14 +7,11 @@ description: Ecommerce Additional Product Code Registration Credential
...
   type:
-    oneOf:
-      - type: array
-        items:
-          type: string
-          enum:
-            - AdditionalProductCodeRegistrationCredential
-      - type: string
-        const: AdditionalProductCodeRegistrationCredential
+    type: array
```
mprorock, line 10: Double typed here?

mprorock (approved):

mprorock:
@rhofvendahl this is a good approach - we should do similar to items we are working on

docs/openapi/components/schemas/common/AdditionalProductCodeRegistrationCredential.yml
nissimsan: @mprorock, yes, the type element is an array type.

BenjaminMoe (approved):

mkhraisha (approved):

PR #535, opened 2022-08-17
Mavennet update organization name to name in postal address 

mkhraisha:
Change organizationName to name in the PostalAddress Object. The organizationName is not generic enough to represent all objects that use Postal Address object. They are not necessarely organizations as they can represent any entity. For example, Schema.org uses name inside the Postal Address schema.

Closing #533


OR13 (approved):

nissimsan (approved):

nissimsan:
This deserves a `merge-first`.

BenjaminMoe (approved):

Issue #527, opened 2022-08-16
How to handle oneOf in the JSON Schema to JSON-LD context code


BenjaminMoe:
Update JSON-LD processor to support oneOf schema functionality

nissimsan:
What should the resulting @context look like to support this?

OR13:
On the call, we discussed how `domain` and `range` might be relevant to `oneOf` and `anyOf`.

cc @TallTed

nissimsan:
@TallTed , please add requirements how we can solve this

TallTed:
@TallTed to consider and rant about rdfs:domain, rdfs:range vs schema:domainIncludes, schema:rangeIncludes and oneOf, anyOf ...

TallTed:
(rant posted above)

TallTed:
It would have been super cool (and still could be!) to include a link to [the issue that prompted this one](https://github.com/w3c-ccg/traceability-vocab/issues/395) in the initial comment on this issue.

(The immediate "@[BenjaminMoe](https://github.com/BenjaminMoe) BenjaminMoe mentioned this issue [on Aug 16](https://github.com/w3c-ccg/traceability-vocab/issues/527#ref-issue-1216458166)" here, that linked to #395, does not achieve the same result, as there's no indication here of *WHY* this issue was referenced over there.)

BenjaminMoe:
In what cases or scenarios would we want to use `oneOf`, are there any requirements around this?

BenjaminMoe:
I think this issue is a result of confusion with the way JSON schema and JSON-LD are displayed with syntax. To do a minimum reproduction of the source for this issue, we can look at the following example:

```
oneOf:
  leiCode:
    title: Legal Entity Identifier Code
    description: Legal entity identifier code
    type: string
    $linkedData:
      term: leiCode
      '@id': https://schema.org/leiCode
  globalLocationNumber:
    title: Global Location Number
    description: The GS1 GLN of the organization.
    type: string
    $linkedData:
      term: globalLocationNumber
      '@id': https://schema.org/globalLocationNumber
```

In this example we have the case where you need to enter either a `leiCode` or `globalLocationNumber`. Where you need one or the other. The linked data terms are added as attributes under each one of these. So we have the question of how JSON-LD would handle this optionality.

The answer is that the schema is effectively orthogonal to the way JSON-LD does processing. So in this example we would expect the result of the JSON-LD context to have both terms defined.

```
{
[...]

PR #539, opened 2022-08-23
Port openapi-to-context.js to editable script

BenjaminMoe:
Right now our openapi to context script is calling a library from `@transmute/jsonld-schema`. This PR updates our script to run directly in the repository as opposed to having to edit an external library to make changes to how the `@context` file is produced.

### DIFF
--- a/packages/traceability-schemas/scripts/openapi-to-context.js
+++ b/packages/traceability-schemas/scripts/openapi-to-context.js
@@ -1,5 +1,6 @@
 /* eslint-disable implicit-arrow-linebreak */
 /* eslint-disable operator-linebreak */
+
 const fs = require('fs');
 const path = require('path');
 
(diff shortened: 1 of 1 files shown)


BenjaminMoe:
The script produces a context file that is very close but not identical to the current functionality. I'm currently looking into the differences and the reasons for them. 

**1. CargoItem missing terms**

Screenshot 2022-08-23 at 16-00-24 Diffchecker

**2. ConsignmentItem missing from existing context**

Screenshot 2022-08-23 at 16-09-27 Diffchecker

BenjaminMoe:
This PR is preparation for addressing: https://github.com/w3c-ccg/traceability-vocab/issues/527

nissimsan (approved):

BenjaminMoe:
Original has `description` in `DCSAShippingInstruction`, which is weird because https://github.com/w3c-ccg/traceability-vocab/blob/main/docs/openapi/components/schemas/common/DCSAShippingInstruction.yml doesn't define `description` in the file. 

Screenshot 2022-08-30 at 22-52-06 Diffchecker

Original is missing RDF class for `Instructions`, which is also weird because we have a term for it: https://github.com/w3c-ccg/traceability-vocab/blob/main/docs/openapi/components/schemas/common/Instructions.yml

Screenshot 2022-08-30 at 22-52-28 Diffchecker

mkhraisha (approved):

OR13 (approved):

Issue #385, opened 2022-04-20
EPCIS traceability alignment


nissimsan:
@mgh128, I just got a link to your white paper from Phil: 
https://gs1.github.io/EndToEndTraceability/GS1-WhitePaper-VerifiableCredentials-EPCIS-end-to-end-traceability.pdf
Some of the terms seem very similar to event types which we recently added to this repo: [TransformEvent](https://github.com/w3c-ccg/traceability-vocab/blob/main/docs/openapi/components/schemas/common/TransformEvent.yml), [TransferEvent](https://github.com/w3c-ccg/traceability-vocab/blob/main/docs/openapi/components/schemas/common/TransferEvent.yml), and a few others. It'd be great if you can take a quick look from a harmonization point of view. Specifically, we shouldn't be re-invent something like `https://w3id.org/traceability#TransformEvent` if there is already a GS1 definition.

mgh128:
Hi @nissimsan 
Thanks for the interest.  The GS1 EPCIS standard ( http://gs1.org/epcis ) was first ratified and published in 2007 and currently (as of v1.2) supports four event types:

- ObjectEvent
- AggregationEvent
- TransactionEvent
- TransformationEvent

v2.0 (the result of almost four years of major modernisation efforts, including the addition of JSON/JSON-LD data format, REST / OpenAPI interface, business-relevant sensor data and various other improvements, including persistentDisposition and formal Linked Data ontologies for EPCIS and CBV) also introduces a further standardised event type (AssociationEvent) for more permanent associations (such as the attachment of a sensor device to a returnable transport container such as a plastic crate, such that the sensor device remains attached even when the fruit/vegetables are emptied from the crate).  You're very welcome to look at https://github.com/gs1/EPCIS to see our current draft examples and schema. Publication of EPCIS/CBV v2.0 and its schema and artefacts is expected within Q2 of 2022.  We're currently doing final corrections and checks.

I took a quick look at [TransformEvent](https://github.com/w3c-ccg/traceability-vocab/blob/main/docs/openapi/components/schemas/common/TransformEvent.yml), [TransferEvent](https://github.com/w3c-ccg/traceability-vocab/blob/main/docs/openapi/components/schemas/common/TransferEvent.yml) and I agree that there is some potential overlap.

[...]

PR #552, opened 2022-08-30
GS1 EPCIS Events

nissimsan:
This is a basic proof of concept, implementing the GS1 EPCIS event model with Verifiable Credentials based on the trace vocab infrastructure. 

@mgh128, @CraigRe, @VladimirAlexiev I'd love your initial thoughts on this? Just general direction - no need to point out missing details. (I just focused on the most basic elements and also consciously skipped the epcisBody).

The resulting VC can be verified with DID Actor (link below). Note that it will fail until the PR is merged which updated the traceability context. 

https://api.did.actor/v/eJylUluTmjAY_S_pq5dw0VWeuou4ahXvl7WzsxNCxCAQNgla3fG_N6hV61unDA8kOTnfufAFvmOWSPJLAusnWEuZCqtc3u12pZ1RYjwo61CrlTEnPkkkRZEobzVQuAEN6p9gkiNMkEcjKvc55L0A5D4lOemMcLqiyIuIfaVRFM7Abo8bDGex2rIJlwqEkST5VSpEhhJMGvnaAjrU9SKsFQ040aAF83cJzijCgfV1HdXnAUroAUnKkhOPr2771Lc2ZG8dqr2NbA2ZsRqN2Hpjmr7wnKeqndFJ-hLve_NPfTYLt7MqehkTx5iCYwHcjI8zLyRY3k_7y0E-jmzVR5eKPMsH2LMQDNOTMGd7gXv00GX4LPaOdhCpKK_qM55YJMWWMiGCKLFMqOmGWSlB9eglqGIIIuah6A-Tm8VeHgp4BB5PA8eSpOoQCUFiL6JJAI7v6iTljK1uIoDj65WKVh_TIEEy4yT_C8ApDtWI_1CJZhmmBc28ku2p67OSHpFr9j8FfPvnxi5GBhlPmSAXn_xOTAGEO6H2yb6z9l4x7dNOc-mMJsNxW7Tjtu7a7eoybgqsT9Xa3aPFkPYjQd_CN9iOtHqp1Fxko5Q4C-x-eOFHEMOVGw1r3Y6sNzY_nka9xmGaBk3HpMX5Wm-G2_7M5a1dyAeB1-sVG-h11Di4AZzONqxVS5KlUZkv7WdwPP4GZaMpqw

https://github.com/w3c-ccg/traceability-vocab/issues/385

### DIFF
--- a/docs/openapi/components/schemas/common/EPCISAssociationEvent.yml
+++ b/docs/openapi/components/schemas/common/EPCISAssociationEvent.yml
@@ -1,0 +1,39 @@
+$linkedData:
+  term: EPCISAssociationEvent
+  '@id': https://ref.gs1.org/epcis/AssociationEvent
+title: EPCIS Association Event
+description: The event type AssociationEvent describes the association or disassociation of one or several physical objects with a parent object or a specific physical location. Like the AggregationEvent, the AssociationEvent is also used to capture associations where there is a strong physical relationship between the containing and the contained objects such that they will all occupy the same location at the same time, until such time as they are disaggregated. However, the AggregationEvent does not allow for associations of objects with physical locations; if action is DELETE while omitting the childEPC and childQuantityList field, all contained children are disaggregated from the containing parent. Because there are situations in which associations are more permanent, i.e. beyond the physical flow of goods (e.g. packing/unpacking and loading/unloading), an AssociationEvent SHOULD be used (a) when objects need to be associated with a physical location or (b) when the parent object could also be subject to other, more temporary associations (i.e. captured using AggregationEvent).
+type: object
+properties:
+  type:
+    oneOf:
+      - type: array
+        items:
+          type: string
+          enum:
+            - EPCISAssociationEvent
+  bizLocation:
+    title: Business Location
+    description: The business location where the objects associated with the EPCs may be found, until contradicted by a subsequent event.
+    $ref: ./Place.yml
+    $linkedData:
+      term: bizLocation
+      '@id': https://ref.gs1.org/epcis/bizLocation
+  bizStep:
+    title: Business Step
+    description: The business step of which this event was a part.
+    type: string
+    $linkedData:
+      term: bizStep
+      '@id': https://ref.gs1.org/epcis/bizStep
+additionalProperties: false
+example: |-
+  {
+    "type": ["EPCISAssociationEvent"],
+    "bizLocation": {
+      "type": ["Place"],
+      "id": "urn:epc:id:sgln:4012345.00002.0",
+      "globalLocationNumber": "4012345.00002.0"
+    },
+    "bizStep": "assembling"
+  }
--- a/docs/openapi/components/schemas/credentials/EPCISDocumentCertificate.yml
+++ b/docs/openapi/components/schemas/credentials/EPCISDocumentCertificate.yml
@@ -1,0 +1,87 @@
+$linkedData:
+  term: EPCISDocumentCertificate
+  '@id': https://w3id.org/traceability#EPCISDocumentCertificate
+title: Commercial Invoice Certificate
+tags:
+  - eCommerce
+description: >-
+  The document recording a transaction between the seller and the buyer. Commercial invoices are normally prepared by sellers.
+  The commercial invoice on itself does not grant any ownership of the goods, unless it has an attached document proving the importer's payment for the total amount. The number of copies of the invoice (both original and copies) required for the delivery of the goods, must be agreed with the importer. Usually, invoices are issued with the original and two copies. Although normally the legislation in different countries does not limit the number of originals, it is not advisable to make more than those strictly necessary in order to accomplish with the customs needs required by the buyer. It is advisable that the importer confirms with the exporter all data that the invoice must provide before its issuing, as well as the particularities it must include in order to accomplish with the regulation of the destination country.
+  (source: Olegario Llamazares: Dictionary Of International Trade, Key definitions of 2000 trade terms and acronyms).
+type: object
+properties:
+  '@context':
+    type: array
+    items:
+      type: string
+      enum:
+        - https://www.w3.org/2018/credentials/v1
+        - https://w3id.org/traceability/v1
+  type:
+    type: array
+    items:
+      type: string
+      enum:
+        - VerifiableCredential
+        - EPCISDocumentCertificate
+  id:
+    type: string
+  name:
+    type: string
+  description:
+    type: string
+  issuanceDate:
+    type: string
+  issuer:
+    $ref: ../common/Organization.yml
+  credentialSubject:
+    $ref: ../common/EPCISDocument.yml
+  proof:
+    type: object
+additionalProperties: false
+required: []
+example: |-
+  {
+    "@context": [
+      "https://www.w3.org/2018/credentials/v1",
+      "https://w3id.org/traceability/v1"
+    ],
+    "type": [
+      "VerifiableCredential",
+      "EPCISDocumentCertificate"
+    ],
+    "issuanceDate": "2022-08-30T10:00:00Z",
+    "issuer": {
+      "type": [
+        "Organization"
+      ],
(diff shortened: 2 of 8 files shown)


docs/openapi/components/schemas/common/EPCISAssociationEvent.yml
```diff
@@ -0,0 +1,39 @@
...
+  '@id': https://ref.gs1.org/epcis/AssociationEvent
+title: EPCIS Association Event
+description: The event type AssociationEvent describes the association or disassociation of one or several physical objects with a parent object or a specific physical location. Like the AggregationEvent, the AssociationEvent is also used to capture associations where there is a strong physical relationship between the containing and the contained objects such that they will all occupy the same location at the same time, until such time as they are disaggregated. However, the AggregationEvent does not allow for associations of objects with physical locations; if action is DELETE while omitting the childEPC and childQuantityList field, all contained children are disaggregated from the containing parent. Because there are situations in which associations are more permanent, i.e. beyond the physical flow of goods (e.g. packing/unpacking and loading/unloading), an AssociationEvent SHOULD be used (a) when objects need to be associated with a physical location or (b) when the parent object could also be subject to other, more temporary associations (i.e. captured using AggregationEvent).
```
TallTed, line 5:
```suggestion
description: The event type AssociationEvent describes the association or disassociation of one or several physical objects with a parent object or a specific physical location. Like the AggregationEvent, the AssociationEvent is used to capture associations where there is a strong physical relationship between the containing and the contained objects, such that they will all occupy the same location at the same time, until such time as they are disaggregated. However, the AggregationEvent does not allow for associations of objects with physical locations; if action is DELETE while omitting the childEPC and childQuantityList field, all contained children are disaggregated from the containing parent. Because there are situations in which associations are more permanent, i.e., beyond the physical flow of goods (e.g., packing/unpacking and loading/unloading), an AssociationEvent SHOULD be used (a) when objects need to be associated with a physical location or (b) when the parent object could also be subject to other, more temporary associations (i.e., captured using AggregationEvent).
```

docs/openapi/components/schemas/credentials/EPCISDocumentCertificate.yml
```diff
@@ -0,0 +1,87 @@
+$linkedData:
+  term: EPCISDocumentCertificate
+  '@id': https://w3id.org/traceability#EPCISDocumentCertificate
+title: Commercial Invoice Certificate
+tags:
+  - eCommerce
+description: >-
+  The document recording a transaction between the seller and the buyer. Commercial invoices are normally prepared by sellers.
+  The commercial invoice on itself does not grant any ownership of the goods, unless it has an attached document proving the importer's payment for the total amount. The number of copies of the invoice (both original and copies) required for the delivery of the goods, must be agreed with the importer. Usually, invoices are issued with the original and two copies. Although normally the legislation in different countries does not limit the number of originals, it is not advisable to make more than those strictly necessary in order to accomplish with the customs needs required by the buyer. It is advisable that the importer confirms with the exporter all data that the invoice must provide before its issuing, as well as the particularities it must include in order to accomplish with the regulation of the destination country.
```
TallTed, line 9:
```suggestion
  In itself, the commercial invoice does not grant any ownership of the goods, unless it has an attached document proving the importer's payment for the total amount. The number of copies of the invoice (both original and copies) required for the delivery of the goods, must be agreed with the importer. Usually, invoices are issued with the original and two copies. Although normally the legislation in different countries does not limit the number of originals, it is not advisable to make more than those strictly necessary in order to comply with the customs needs required by the buyer. It is advisable that the importer confirm with the exporter all data that the invoice must provide before its issuing, as well as the particularities it must include in order to comply with the regulation of the destination country.
```

mgh128:
Thanks for preparing this.  I clicked on the link and although I definitely don't want to discourage you, I have a few comments:

The part that somewhat resembles an EPCIS event won't validate against our schema at https://ref.gs1.org/standards/epcis/epcis-json-schema.json  or
https://ref.gs1.org/standards/epcis/epcis-shacl.ttl

because 

- `eventList` is not expected as a direct property of `EPCISDocument`
- `eventTime` is missing from the event - so there's no indication of the timestamp of the event
- Neither a `parentID` nor `childQuantityList` or `childEPCs` are present - so there is no indication of which objects were involved
- `"bizStep" : "assembling"`  only works if the standard JSON-LD context for EPCIS is referenced (at https://ref.gs1.org/standards/epcis/epcis-context.jsonld ), since this maps the bare word "assembling" to the CURIE `cbv:BizStep-assembling` (which in turn corresponds to https://ref.gs1.org/cbv/BizStep-assembling - and that URL should resolve directly to the term as soon we correct that temporary glitch when somebody changed the server configuration during July/August, even though it had been working previously). 
- Also, both `schema:globalLocationNumber` and `gs1:globalLocationNumber` expect a 13-digit numeric string representation of a GLN - without any dots or any of the other weird re-arrangement that the EPC URN formats use.  (Thankfully, a constrained subset of GS1 Digital Link URIs can now be used in EPCIS / CBV 2.0 instead of EPC URNs, while the recently ratified GS1 Tag Data Standard 2.0 introduces new EPC binary encodings that avoid the rearrangement and also translate in a more straightforward way to/from GS1 Digital Link URIs or element strings and eschew the EPC URN notation altogether for those new EPC schemes)

If it's easier to take an example of an AssociationEvent that should validate, please consider using something such as https://ref.gs1.org/docs/epcis/examples/association_event-a.jsonld  - it's already fairly minimal

nissimsan:
@mgh128,

Excellent feedback. Not discouraged at all! :) Your first 4 points I 100% expected. Like I said, I only aimed for a super basic starting point, so all those points were expected. So thanks for confirming!

I'm not exactly sure how to deal with bullet 4. Will need some nerding around. Any input on how to make it work with the trace-vocab constraints is appreciated. 

Your last bullet we also discussed on the trace call yesterday. We have a `Place` class already which also has a GLN, and was hinting at using it. But it isn't the exact same, so probably won't be compatible that way.

mgh128:
@nissimsan 

Regarding bullet point 4, did you already try referencing (within your existing `@context` declaration) our standard JSON-LD context for EPCIS ( https://ref.gs1.org/standards/epcis/epcis-context.jsonld )  - and if so, did it break anything?  If not, then I'd expect that it addresses bullet point 4.

Regarding the last bullet point, https://ref.gs1.org/epcis/bizLocation already has a [rdfs:range](http://www.w3.org/2000/01/rdf-schema#range) of [gs1:Place ](https://www.gs1.org/voc/Place) (which is modelled on [schema:Place](https://schema.org/Place) but extends it in some ways).  If instead of using a URN (URI that typically Resolves Nowhere) such as  `urn:epc:id:sgln:4012345.00002.0` you instead use a corresponding [GS1 Digital Link](https://www.gs1.org/standards/gs1-digital-link) URI such as https://id.gs1.org/414/4012345000023  then it can much more easily link to any Linked Data about that [gs1:Place ](https://www.gs1.org/voc/Place) or [schema:Place](https://schema.org/Place) .  

Typically within EPCIS event data, we avoid bloating the event data by minimising the amount of unnecessary 'master data' / static data and instead just use URIs, assuming that if anyone needs the master data, they can look it up via the URI.  Obviously that's much more practical with a Web URI than a URN.  So we wouldn't typically include [schema:globalLocationNumber ](https://schema.org/globalLocationNumber)or [gs1:globalLocationNumber](https://www.gs1.org/voc/globalLocationNumber) within the event data, nor any other descriptive details that are the same each time that URI appears within an event.  

In case you're wondering what /414/ is within https://id.gs1.org/414/4012345000023 it's the [GS1 Application Identifier](https://www.gs1.org/standards/barcodes/application-identifiers?lang=en)  for a GS1 Global Location Number of a physical location.   (414 = physical location GLN, 417 = GLN of an organisation)

nissimsan:
Closing this Draft PR - I will move this discussion to an issue.

PR #563, opened 2022-09-16
Added missing mech properties

nissimsan:
MTR was missing mech property examples


docs/openapi/components/schemas/credentials/MillTestReportCertificate.yml
```diff
@@ -478,8 +478,8 @@ example: |-
                 "type": [
                   "MechanicalProperty"
                 ],
-                "identifier": "",
-                "name": "Gauge Length"
+                "identifier": "ISO 1352:2021",
```
mkhraisha, line 481: https://github.com/w3c-ccg/traceability-vocab/blob/e92bb49a71b74c3d05fe75dc76ef1e99fdde3d80/docs/openapi/components/schemas/common/SteelProduct.yml#L119
These seem to have a minor difference in them, probably a good idea for them to be identical

docs/openapi/components/schemas/credentials/MillTestReportCertificate.yml
```diff
@@ -497,7 +497,7 @@ example: |-
                 "type": [
                   "MechanicalProperty"
                 ],
-                "identifier": "",
+                "identifier": "ISO 1519:2011",
```
mkhraisha, line 500: This one does not exist on the steel product

mkhraisha (changes requested):

brownoxford:
Consensus on call to fix and merge when ready

nissimsan:
I'll remove the package lock before merging

nissimsan:
@mkhraisha, I've copied the examples from SteelProduct now instead as you suggest. Made a separate PR to avoid git acrobatics: https://github.com/w3c-ccg/traceability-vocab/pull/580

PR #608, opened 2022-10-31
Create schema for SC-237A form

rhofvendahl:
This PR creates a schema for an  SC-237A form.

References:
https://www.ams.usda.gov/resources/sc237a
https://www.ams.usda.gov/sites/default/files/media/SC237A.pdf

### DIFF
--- a/docs/openapi/components/schemas/common/SC237AForm.yml
+++ b/docs/openapi/components/schemas/common/SC237AForm.yml
@@ -1,0 +1,189 @@
+$linkedData:
+  term: SC237AForm
+  '@id': https://w3id.org/traceability#SC237AForm
+title: SC-237A Form
+description: Information regarding an application for audit services.
+type: object
+properties:
+  type:
+    oneOf:
+      - type: array
+        items:
+          type: string
+          enum:
+            - SC237AForm
+      - type: string
+        const: SC237AForm
+  requestDate:
+    title: Request Date
+    description: Date the audit was requested.
+    type: string
+    $linkedData:
+      term: requestDate
+      '@id': https://service.unece.org/trade/uncefact/vocabulary/uncefact/#reportSubmissionDateTime
+  anticipatedAuditDate:
+    title: Anticipated Audit Date
+    description: Anticipated date of the audit.
+    type: string
+    $linkedData:
+      term: anticipatedAuditDate
+      '@id': https://www.gs1.org/voc/certificationAuditDate
+  auditee:
+    title: Auditee
+    description: The party to be audited.
+    $ref: ./Organization.yml
+    $linkedData:
+      term: auditee
+      '@id': https://service.unece.org/trade/uncefact/vocabulary/uncefact/#associatedParty
+  applicant:
+    title: Applicant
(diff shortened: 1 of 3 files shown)


OR13 (approved):

BenjaminMoe (changes requested):
`SC237AForm` should use the following format for `type`. 

```
    type: array
    readOnly: true
    const:
      - SC237AForm
    default:
      - SC237AForm
    items:
      type: string
      enum:
        - SC237AForm
```

Also add `required` after `additionalProperties: false`
```
required:
  - type
```

docs/openapi/components/schemas/common/SC237AForm.yml
```diff
@@ -0,0 +1,189 @@
+$linkedData:
+  term: SC237AForm
+  '@id': https://w3id.org/traceability#SC237AForm
+title: SC-237A Form
+description: Information regarding an application for audit services.
```
mkhraisha, line 5: Can we link to a hosted version of this form from here? We're doing similar for other forms.

rhofvendahl: @mkhraisha 

Do you mean the pdf? As in https://www.ams.usda.gov/sites/default/files/media/SC237A.pdf ?

I'm concerned that a link like that might break, but a link to the USDA page for that form (https://www.ams.usda.gov/resources/sc237a) should work, as it references the form and is more likely to be updated.

I'm open to doing that, but would it work instead to just have a more detailed description (without breakable links)? I've been meaning to do that anyway.

BenjaminMoe (approved):

docs/openapi/components/schemas/common/SC237AForm.yml
```diff
@@ -0,0 +1,193 @@
+$linkedData:
+  term: SC237AForm
+  '@id': https://w3id.org/traceability#SC237AForm
+title: SC-237A Form
```
nissimsan, line 4: Two suggestions: 
1. Please prefix with `USDA`
2. Spell out "SC" to make this less esoteric. 

I would suggest naming this `USDASpecialtyCrops237AForm`.

nissimsan (changes requested):

docs/openapi/components/schemas/common/SC237AForm.yml
rhofvendahl: @nissimsan I just made this change, could you mark this request as complete so it doesn't block the merge?

nissimsan: Excellent

nissimsan (approved):

docs/openapi/components/schemas/common/SC237AForm.yml
mkhraisha: yes a detailed description instead works.

mkhraisha (changes requested):
Please update the description

rhofvendahl:
@mkhraisha I've updated the description for clarity, should be ready to merge.

OR13 (approved):

PR #617, opened 2022-11-10
Add schema type to issuer

BenjaminMoe:
We don't have a defined type for `issuer`, which is resulting in a lot of undefined terms that are not being checked. This PR adds a type to the issuer to make sure everything has a defined schema.


nissimsan (approved):

nissimsan:
We need another approval, pls

OR13 (approved):

PR #623, opened 2022-11-12
Add schemas for PPQ 449-R

rhofvendahl:
This PR creates schemas for PPQ 449-R, Temperature Recording Calibration Report.


nissimsan (approved):

nissimsan:
Hmm, wondering if `Temperature Recording` is not a type of `Measurement`?

rhofvendahl:
I think `Observation` (with props `measurement` and `property`) would be inconvenient for this use case, at least as it is now.

Each temperature reading must have a bulb number and a value representing the instrument reading at an actual temp of 0deg C. However, `MeasuredProperty` is currently only suited for mechanical or chemical properties, and additionally wouldn't support having both property values.

In the future I'm hoping to update Observation or make an alternative that'd be flexible enough to cover some of these inspection cases.

PR #630, opened 2022-11-22
feat: ctpat expiration and status

nissimsan:
Adds credentialStatus
Makes expirationDate mandatory


nissimsan:
I fixed this context issue locally by adding a copy of https://w3id.org/vc/status-list/2021/v1. But that folder is gitignored, so not sure how to get this passing.

PR #639, opened 2022-11-29
Create gap location certification schema

rhofvendahl:
This PR creates a schema to allow tracking & revocation of GAP certifications.


mkhraisha (approved):
Assuming Nis's  comment about acronyms is addressed lgtm

Issue #624, opened 2022-11-12
Proposal: Update Tooling to separate Schema from Context


BenjaminMoe:
## Description of Problem:

Right now we have JSON Schema and Context definitions in the same file. From working with this approach I think this creates some of the following problems.

1. **Bloat**. Anytime you use organization, you pull in things like LEI code or IATA code, which are required for specific credentials related to organizations, but not applicable to most (eg. "what i need an iata code to apply for an Organic Certification?). 
2. **Confusion** between wrapper and associated documents and RDF classes. `Bill of Lading Credential` outer wrapper versus the `Bill of Lading` schema used to defined the `credentialSubject`. 
3. **Confusion** with respect to the `$linkedData` element on our schema yml files with respect to `@id` and `@type`. 

## Proposal For Changes:

Summary Separate `@context` from JSON Schema. 

1. We have a directory of JSON files that define RDF classes that get built into a single `@context` file
2. We have a single directory for credential schema definitions
   - Schema definitions can reference other components for convenience
 
## Advantages of this Approach:

1. Addresses: https://github.com/w3c-ccg/traceability-vocab/issues/527
2. Allows for document specific requirements for optional and required fields

### How does this address Issue 527 (oneOf)

Given the following example context

```JSON
{
  "@context": {
    "@version": 1.1,
    "Organization": {
      "@id": "https://schema.org/Organization",
      "@context": {
        "name": {
          "@id": "https://schema.org/name"
        }
      }
    },
    "Person": {
      "@id": "https://schema.org/Person",
      "@context": {
        "name": {
          "@id": "https://schema.org/name"
        }
      }
    }
  }
}
```
Would allow you to define JSON context similar to the following:

```yml
title: Issuer
description: An issuer such as a person or organization
type: object
properties:
  type:
    type: array
    items:
      type: string
      enum:
        - Organization
        - Person
  id:
    title: Identifier
    description: Decentralized identifier for the issuer
    type: string
  name:
    title: Name
    description: Name of the issuer.
    type: string
additionalProperties: false
required:
  - id
  - type
  - name
example: |-
  {
    "id": "did:key:z6MktHQo3fRRohk44dsbE76CuiTpBmyMWq2VVjvV6aBSeE3U",
    "type": ["Person"],
    "name": "Glover - Gleason"
  }
```

[...]

PR #644, opened 2022-11-29
Narrowly define schema for commercial invoice

BenjaminMoe:
For the commercial invoice certificate, we are including `exporter`, `importer`, `shipFromParty`, `shipToParty`, `provider`. Which provides too much flexibility for a single credential type. To accommodate these situations we can define more invoice types.

### DIFF
--- a/docs/openapi/components/schemas/credentials/CommercialInvoiceCredential.yml
+++ b/docs/openapi/components/schemas/credentials/CommercialInvoiceCredential.yml
@@ -95,7 +95,278 @@ properties:
       - type
       - name
   credentialSubject:
-    $ref: ../common/Invoice.yml
+    title: Invoice
+    description: A statement of the money due for goods or services; a bill.
+    type: object
+    properties:
+      type:
+        type: array
+        readOnly: true
+        const:
+          - Invoice
+        default:
+          - Invoice
+        items:
+          type: string
+          enum:
+            - Invoice
+      identifier:
+        title: Property Identifier
+        description: Identifiers for a property.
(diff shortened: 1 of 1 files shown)


docs/openapi/components/schemas/credentials/CommercialInvoiceCredential.yml
```diff
@@ -95,7 +95,278 @@ properties:
       - type
       - name
   credentialSubject:
-    $ref: ../common/Invoice.yml
+    title: Invoice
```
mkhraisha, line 98: Why are you defining this inline instead of modifying the existing common object? 
Doing it this way will reduce reusability.

mkhraisha (changes requested):

docs/openapi/components/schemas/credentials/CommercialInvoiceCredential.yml
BenjaminMoe: This is with respect to issue: https://github.com/w3c-ccg/traceability-vocab/issues/624. Pulling in the entire object can pull in a lot of attributes that the issuer is not interested in. This narrows down the scope to only use the attributes required by the document the credential is replacing.

nissimsan (approved):

jmmski (approved):

nissimsan:
@mkhraisha let's discuss this offline. Agreement on the call that we can merge this if you're okay with this way of narrowing the definitions.

mkhraisha (approved):

PR #645, opened 2022-11-29
MTR schema narrowing

nissimsan:
Removal of unnecessary parts of MTR.

### DIFF
--- a/docs/openapi/components/schemas/credentials/MillTestReportCredential.yml
+++ b/docs/openapi/components/schemas/credentials/MillTestReportCredential.yml
@@ -134,11 +134,53 @@ properties:
           - name
           - phoneNumber
           - location
-      # Schema Definition
       product:
-        $ref: ../common/SteelProduct.yml
-        title: Product
-        description: Product details
+        title: Steel Product
+        description: A finished or unfinished manufactured steel product
+        type: object
+        properties:
+          type:
+            type: array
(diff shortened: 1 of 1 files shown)


mkhraisha (changes requested):
Why are you removing the dependence on steel product and defining it inline? If there are alterations to the base steel product why aren't they being tracked on that common object?

BenjaminMoe (approved):

jmmski (approved):

nissimsan:
@mkhraisha, copying this from DM...
You are asking about the way we are narrowing the schemas on https://github.com/w3c-ccg/traceability-vocab/pull/644 and https://github.com/w3c-ccg/traceability-vocab/pull/645. We’ve started doing that recently as a way to chip away at what gets included from the final JSON schema.
The LD terms still only get the defined on the RDF class. But on the JSON Schema side, instead of always pulling everything into each credential, we have started picking just what is needed.
For example, currently the MTR literally includes everything from Invoice (because of a connection which has been added for other credentials but have no relevance to the MTR). That makes the resulting schema very large and difficult for end user to navigate. So PR645 removes all that invoice stuff from the MTR. It just doesn’t belong there.
I’d like to ask you to please approve those two PRs. We discussed this on the call , merge is only pending you changing to approve.

mkhraisha (approved):

PR #649, opened 2022-12-02
GS1 license credentials

nissimsan:
https://ref.gs1.org/gs1/vc/data-model/ 
- GS1PrefixLicenceCredential
- GS1CompanyPrefixLicenceCredential

### DIFF
--- a/docs/openapi/components/schemas/credentials/GS1CompanyPrefixLicenseCredential.yml
+++ b/docs/openapi/components/schemas/credentials/GS1CompanyPrefixLicenseCredential.yml
@@ -1,0 +1,107 @@
+$linkedData:
+  term: GS1CompanyPrefixLicenceCredential
+  '@id': https://w3id.org/traceability#GS1CompanyPrefixLicenceCredential
+title: GS1 Company Prefix Licence Credential
+tags:
+  - eCommerce
+description: >-
+  A GS1 Company Prefix Licence is issued by a GS1 Member Organisation or GS1 Global Office and allocated to a user company or to itself for the purpose of generating tier 1 GS1 identification keys.
+type: object
+properties:
+  '@context':
+    type: array
+    readOnly: true
+    const:
+      - https://www.w3.org/2018/credentials/v1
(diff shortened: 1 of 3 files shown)


BenjaminMoe (approved):

docs/openapi/components/schemas/credentials/GS1CompanyPrefixLicenseCredential.yml
```diff
@@ -0,0 +1,107 @@
+$linkedData:
+  term: GS1CompanyPrefixLicenceCredential
+  '@id': https://w3id.org/traceability#GS1CompanyPrefixLicenceCredential
+title: GS1 Company Prefix Licence Credential
```
TallTed, line 4: Is this meant to be US English or UK English?  Note that W3 intends to use US ...

If there's room for a comment somewhere, it might be good to note when UK English is intended for something (like perhaps this "Licence"), instead of being perceived as a misspelling (of "License", in this case).

docs/openapi/components/schemas/credentials/GS1CompanyPrefixLicenseCredential.yml
```diff
@@ -0,0 +1,107 @@
...
+    description: Resolvable DID for GS1 Global Office.
+    type: string
+  credentialSubject:
+    type: object
+    properties:
+      id:
+        description: The DID for the user company to whom the GS1 Company Prefix has been licensed.
+        type: string
+      extendsCredential: 
+        description: URI of the GS1 Prefix licence that this GS1 Company Prefix licence extends.
```
TallTed, line 60: and here we blend US and UK spellings...

BenjaminMoe (approved):

docs/openapi/components/schemas/credentials/GS1CompanyPrefixLicenseCredential.yml
TallTed: For certain, we should only use one of these (US or UK) in any given document. 

Optimally, we would only use one of these in any given repo.

Issue #637, opened 2022-11-28
Instances of schemas are non-spanning 


BenjaminMoe:
Add validation check for making sure examples are a spanning checking of their respective schemas. 

**Condition for resolution:**

We have a PR for an added test to see if a credential example includes all described objects. And that each of those objects include all required fields. By definition, and complex object is required.

OR13:
There are examples of credentials where the example contains a subset of the schema.

For example: BillOfLadingCredential

We consider every instance (example) that contains a subset of the schema (requirements) to be an error.

Our current tests in this repo don't catch these errors, but our tests outside this repo do.

Each example must contain ALL fields that are required per the schema definition, and the schema requirements should be negotiated to ensure this is not burdensome.

For example:

issuer: { type: Organization }

This might imply { type: Organization, location: { type: Place, geo: { type : GeoCoordinates } } }

The JSON Schema, NOT the JSON-LD context, determines what a "valid credential is", because validation is a function of the instance data, and the schema.

If the schema contains `geo` the example MUST contain `geo`.
If the schema does not contain `geo` the example MUST NOT contain `geo`.


## Actions:

We need a way to test to ensure our examples are "good" and that they "match the required fields of the schemas".

If there is a mismatch, the trivial solution is to update the example first, then file a new issue on this repo, requesting changes to the schema second.

See for example:

https://github.com/transmute-industries/vc-credential-schema-open-api-specification

nissimsan:
Verifiable Business Card has been narrowed down, it should serve as an example of a "full coverage" example.

nissimsan:
Ping, @BenjaminMoe

nissimsan:
Work ahead: 
- v2 examples always valid
- Schema attribute

This will automatically handled in the v2 test suite.

OR13:
suggest closing

PR #650, opened 2022-12-06
Fix InspectionReport.yml syntax

BenjaminMoe:
For `inspectors` an array was being defined without defining any items. Previously wasn't caught because our examples are a subset of our schemas. https://github.com/w3c-ccg/traceability-vocab/issues/637


nissimsan (approved):

mkhraisha (approved):
Good catch lgtm

PR #651, opened 2022-12-06
Reverting mandatory expiration

nissimsan:
We declared expiration mandatory a bit prematurely recently. Reverting back until we have the requirements fully settled.


OR13 (approved):
File an issue to "undo this reversion" in the future, since I believe it is correct that these require expiration eventually.

BenjaminMoe (approved):
Issue created: https://github.com/w3c-ccg/traceability-vocab/issues/653

mkhraisha (approved):
Confirming this is only affecting these credentials because of business rules around only these three credentials rather than a desire to state we don't want expiration date in general? @nissimsan if so LGTM

PR #674, opened 2022-12-15
Update Intent to Import

ipbyrne:
This PR is for updating the Intent to Import Credential required fields and examples.

### DIFF
--- a/docs/openapi/components/schemas/credentials/BillOfLadingCredential.yml
+++ b/docs/openapi/components/schemas/credentials/BillOfLadingCredential.yml
@@ -354,9 +354,9 @@ example: |-
     },
     "proof": {
       "type": "Ed25519Signature2018",
-      "created": "2022-12-19T20:08:21Z",
+      "created": "2022-12-20T15:30:19Z",
       "verificationMethod": "did:key:z6MktHQo3fRRohk44dsbE76CuiTpBmyMWq2VVjvV6aBSeE3U#z6MktHQo3fRRohk44dsbE76CuiTpBmyMWq2VVjvV6aBSeE3U",
       "proofPurpose": "assertionMethod",
-      "jws": "eyJhbGciOiJFZERTQSIsImI2NCI6ZmFsc2UsImNyaXQiOlsiYjY0Il19..730GB4RnnbPP_mfUltPn9wPcbPYhSES17UFV50g1qpfeZM4-c0I77womVN4kulqvFszhVn3RP4upFKjsOWjBCw"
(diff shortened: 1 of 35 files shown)


TallTed:
Please update the title of this PR to `Update Intent to Import`.
(Current last word is erroneously `Important`.)

docs/openapi/components/schemas/common/IntentToImport.yml
```diff
@@ -27,15 +27,15 @@ properties:
     description: >-
       Organization that is declaring their intention for a product to be exported to a
       different location than where it was created.
-    type: object
+    $ref: ./Organization.yml
```
BenjaminMoe, line 30: I don't think we want to pull in the whole Organization object. We want to only include a subset of what we want to use. Example here: https://github.com/w3c-ccg/traceability-vocab/blob/intent-to-import/docs/openapi/components/schemas/credentials/MillTestReportCredential.yml#L99

docs/openapi/components/schemas/common/IntentToImport.yml
```diff
@@ -27,15 +27,15 @@ properties:
...
     $linkedData:
       term: exporter
       '@id': https://service.unece.org/trade/uncefact/vocabulary/uncefact/#exporterParty
   importer:
     title: Importer
     description: >-
       Organization that is expected to receive some amount of the product.
-    type: object
+    $ref: ./Organization.yml
```
BenjaminMoe, line 38: Same here, declare an object as a subset of organization.

BenjaminMoe (changes requested):
Request: Narrow definition for organization. You can declare a subset in the `snippets` folder.

BenjaminMoe (approved):

docs/openapi/components/schemas/credentials/BillOfLadingCredential.yml
```diff
@@ -354,9 +354,9 @@ example: |-
     },
     "proof": {
       "type": "Ed25519Signature2018",
-      "created": "2022-11-29T07:37:53Z",
+      "created": "2022-12-20T15:30:19Z",
```
nissimsan, line 357: @ipbyrne, you are re-signing all credentials for no reason. You shouldn't be touching the credentials which aren't actually part of the PR's scope. It makes it hard to understand the actual impact of the PR and causes unnecessary merge conflicts.
Please revert the re-signing of credentials which aren't needed. 
Next time, just run the `regenerate` script, it takes care of all this.

nissimsan (changes requested):

docs/openapi/components/schemas/credentials/BillOfLadingCredential.yml
nissimsan: Beyond the unnecessary resigns, the PR is great

PR #697, opened 2023-01-27
GS1 delegation and data credentials

nissimsan:
Ref https://ref.gs1.org/gs1/vc-data-model/ , sections 4.4.2 and 4.4.3

### DIFF
--- a/docs/openapi/components/schemas/credentials/GS1DataCredential.yml
+++ b/docs/openapi/components/schemas/credentials/GS1DataCredential.yml
@@ -83,7 +83,6 @@ required:
   - issuanceDate
   - issuer
   - credentialSubject
-  - credentialStatus
 example: |-
   {
     "@context": [
--- a/docs/openapi/components/schemas/credentials/GS1DataCredential.yml
+++ b/docs/openapi/components/schemas/credentials/GS1DataCredential.yml
@@ -1,0 +1,126 @@
+$linkedData:
+  term: GS1DataCredential
+  '@id': https://w3id.org/traceability#GS1DataCredential
+title: GS1 Data Credential
+tags:
+  - eCommerce
+description: >-
+  A data credential is data about something identified with a key credential. While anyone can make any declaration about an object, for that data to be trusted, it has to be authorised in some fashion. This can vary depending on the type of data and the business process in which it is used. For example, a brand owner can be trusted to declare the brand name, description, size and unit of measure, and much more about their own product. The brand owner may not be trusted, however, to declare certain product certifications (e.g., Halal and conformance to other religious practices) or to provide high-quality data for critical supply chain processes (e.g., dimensions and weights for planogram purposes). In such cases, the brand owner would have to authorise other parties that are trusted to provide that data.
+  On the other hand, an anonymous consumer can provide a review of the product without any authorisations provided by the brand owner but can increase trust in the review by referencing a proof of purchase credential for the product.
+  Ultimately, the rules governing which declarations may be made, by whom, and with what combination of data authorisation and data certification credentials depend on the data being provided.
+  The key authorisation and data certification credential URIs do not need to be resolvable, and the fact that this credential is presented does not require that the other credentials be presented as well. It is up to the holder to decide whether to present the other credentials alongside this one, and up to the verifier to decide whether or not to accept them (or accept their absence).
+type: object
+properties:
+  '@context':
+    type: array
+    readOnly: true
+    const:
+      - https://www.w3.org/2018/credentials/v1
+      - https://w3id.org/traceability/v1
+      - https://w3id.org/vc/status-list/2021/v1
+    default:
+      - https://www.w3.org/2018/credentials/v1
+      - https://w3id.org/traceability/v1
+      - https://w3id.org/vc/status-list/2021/v1
+    items:
+      type: string
+      enum:
+        - https://www.w3.org/2018/credentials/v1
+        - https://w3id.org/traceability/v1
+        - https://w3id.org/vc/status-list/2021/v1
+  type:
+    type: array
+    readOnly: true
+    const:
+      - VerifiableCredential
+      - GS1DataCredential
+    default:
+      - VerifiableCredential
+      - GS1DataCredential
+    items:
+      type: string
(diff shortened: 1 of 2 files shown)


BenjaminMoe (approved):

docs/openapi/components/schemas/credentials/GS1DataCredential.yml
```diff
@@ -0,0 +1,126 @@
...
+  The key authorisation and data certification credential URIs do not need to be resolvable, and the fact that this credential is presented does not require that the other credentials be presented as well. It is up to the holder to decide whether to present the other credentials alongside this one, and up to the verifier to decide whether or not to accept them (or accept their absence).
+type: object
+properties:
+  '@context':
+    type: array
+    readOnly: true
+    const:
+      - https://www.w3.org/2018/credentials/v1
+      - https://w3id.org/traceability/v1
+      - https://w3id.org/vc/status-list/2021/v1
```
mkhraisha, line 20: Have we updated trace interop to use statuslist2021 instead of revocationlist? That spec says we must use revocationlist2020: https://w3c-ccg.github.io/traceability-interop/draft/#linked-data-signature-suites

Besides what if I want to issue this credential without revocation mechanism? I feel like that restriction belongs on interop not the vocab.

docs/openapi/components/schemas/credentials/GS1DataCredential.yml
```diff
@@ -0,0 +1,126 @@
...
+    default:
+      - VerifiableCredential
+      - GS1DataCredential
+    items:
+      type: string
+      enum:
+        - VerifiableCredential
+        - GS1DataCredential
+  id:
+    description: A DID for this credential. The ID is not resolvable because the credential is sensitive information and only the user company can decide when it is presented.
```
mkhraisha, line 46: does this have to be a DID,? can it not be a normal ID?

docs/openapi/components/schemas/credentials/GS1DataCredential.yml
```diff
@@ -0,0 +1,126 @@
...
+  proof:
+    $ref: ../snippets/proof.yml
+additionalProperties: true
+required:
+  - '@context'
+  - type
+  - issuanceDate
+  - issuer
+  - credentialSubject
+  - credentialStatus
```
mkhraisha, line 86: credentialStatus should not be required, If i wish to issue a non-status tracked credential the vocabulary shouldn't restrict me.

docs/openapi/components/schemas/credentials/GS1DelegationCredential.yml
```diff
@@ -0,0 +1,120 @@
...
+  This is the Verifiable Credential that delegates to another party the right to issue data credentials on behalf of the issuer of this credential.
+type: object
+properties:
+  '@context':
+    type: array
+    readOnly: true
+    const:
+      - https://www.w3.org/2018/credentials/v1
+      - https://w3id.org/traceability/v1
+      - https://w3id.org/vc/status-list/2021/v1
```
mkhraisha, line 17: Similar revocation conversation.

docs/openapi/components/schemas/credentials/GS1DelegationCredential.yml
```diff
@@ -0,0 +1,120 @@
...
+    default:
+      - VerifiableCredential
+      - GS1DelegationCredential
+    items:
+      type: string
+      enum:
+        - VerifiableCredential
+        - GS1DelegationCredential
+  id:
+    description: A DID for this credential. The ID is not resolvable because the credential is sensitive information and only the user company can decide when it is presented.
```
mkhraisha, line 43: similar did/ID comment

mkhraisha (changes requested):

docs/openapi/components/schemas/credentials/GS1DataCredential.yml
nissimsan: Good callout, @mkhraisha. 

As you know, this is copied directly from GS1's data model. I can't change this to 2020 as that would break with the GS1 spec. 

How about avoiding this dependency by letting this pass into trace-vocab as is (just example and context, no specific schemas), and also not include in trace-interop tests before we're ready for 2021?

docs/openapi/components/schemas/credentials/GS1DataCredential.yml
nissimsan: As per the spec, yes. I'll run it by GS1...

nissimsan: PS: Can we agree that DIDs are also "normal"  :)

docs/openapi/components/schemas/credentials/GS1DataCredential.yml
nissimsan: Right! My bad.

docs/openapi/components/schemas/credentials/GS1DelegationCredential.yml
nissimsan: Yep - same answer: I suggest we let this though to avoid a dependency gridlock.

nissimsan:
@mkhraisha, I talked to Paul about the vc.id. I copied the description from the examples section of the gs1 vc spec, so this indeed was not a normative description. There aren't any normative descriptions in the spec, so I just removed those descriptions now. 

Again: good call out! 

Let me know if you're okay on my proposal on dealing with statuslist2021.

docs/openapi/components/schemas/credentials/GS1DataCredential.yml
mkhraisha: Makes sense to me.

mkhraisha (approved):
Lgtm

docs/openapi/components/schemas/credentials/GS1DelegationCredential.yml
OR13: beware that we might need to "update tooling" every-time we "add a new context"....

nissimsan:
Alright, two approvals and some good, resolved conversation - I'll merge this now...

PR #698, opened 2023-01-30
Refactor Organization with external IDs

rhofvendahl:
This PR creates a schema `ExternalIdentifier`, and refactors `Organization` to represent `leiCode`, `globalLocationNumber`, and `taxId` as external ID. This is done in order to keep `Organization` both succinct & flexible in the case of additional IDs which might become necessary.

### DIFF
--- a/docs/openapi/components/schemas/common/AgricultureInspectionGeneric.yml
+++ b/docs/openapi/components/schemas/common/AgricultureInspectionGeneric.yml
@@ -293,7 +293,13 @@ example: |-
       "broker": {
         "type": ["Organization"],
         "name": "Cole United",
-        "leiCode": "54321351219389121979"
+        "externalIdentifiers": [
+          {
+            "type": ["ExternalIdentifier"],
+            "name": "LEI Code",
+            "value": "OHVJ56TBUFWT9XWMBEJ9"
+          }
+        ]
       }
     },
     "inspectionType": "Contamination",
--- a/docs/openapi/components/schemas/common/ExternalIdentifier.yml
+++ b/docs/openapi/components/schemas/common/ExternalIdentifier.yml
@@ -1,0 +1,49 @@
+$linkedData:
+  term: ExternalIdentifier
+  '@id': https://w3id.org/traceability#ExternalIdentifier
+title: External Identifier
+description: Information to recognize & verify an external identifier.
+type: object
+properties:
+  type:
+    type: array
+    readOnly: true
+    const:
+      - ExternalIdentifier
+    default:
+      - ExternalIdentifier
+    items:
+      type: string
+      enum:
+        - ExternalIdentifier
+  name:
+    title: URI
+    description: Name of identifier.
+    type: string
+    $linkedData:
+      term: uri
+      '@id': https://schema.org/name
+  value:
+    title: URI
+    description: Value of identifier.
+    type: string
+    $linkedData:
+      term: uri
+      '@id': https://schema.org/value
+  uri:
+    title: URI
+    description: URI (Universal Resource Identifier) for identifier.
+    type: string
+    $linkedData:
+      term: uri
+      '@id': https://schema.org/url
+additionalProperties: false  
+required:
+  - type
+example: |-
+  {
+    "type": ["ExternalIdentifier"],
+    "name": "IOR_Number",
+    "value": "SOMIDNUM123",
+    "uri": "http://example.gov/ior/"
+  }
--- a/docs/openapi/components/schemas/common/ExternalIdentifier.yml
+++ b/docs/openapi/components/schemas/common/ExternalIdentifier.yml
@@ -43,7 +43,7 @@ required:
 example: |-
   {
     "type": ["ExternalIdentifier"],
-    "name": "IOR_Number",
-    "value": "SOMIDNUM123",
-    "uri": "http://example.gov/ior/"
+    "name": "leiCode",
+    "value": "2594007XIACKNMUAW223",
+    "uri": "http://example.gov/lei/"
   }
--- a/docs/openapi/components/schemas/common/ExternalIdentifier.yml
+++ b/docs/openapi/components/schemas/common/ExternalIdentifier.yml
@@ -40,6 +40,8 @@ properties:
 additionalProperties: false  
 required:
   - type
+  - name
+  - value
 example: |-
   {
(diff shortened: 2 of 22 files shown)


docs/openapi/components/schemas/common/Organization.yml
```diff
@@ -55,13 +48,6 @@ properties:
...
-  globalLocationNumber:
-    title: Global Location Number
-    description: The GS1 GLN of the organization.
-    type: string
-    $linkedData:
-      term: globalLocationNumber
-      '@id': https://schema.org/globalLocationNumber
   location:
     title: Location
     description: The location of, for example, where an event is happening, where an organization is located, or where an action takes place.
```
TallTed, line 53:
```suggestion
    description: The location where, for example, an event happens, an organization is located, or an action takes place.
```

docs/openapi/components/schemas/common/ExternalIdentifier.yml
```diff
@@ -0,0 +1,49 @@
...
+  uri:
+    title: URI
+    description: URI (Universal Resource Identifier) for identifier.
+    type: string
+    $linkedData:
+      term: uri
+      '@id': https://schema.org/url
+additionalProperties: false  
+required:
+  - type
```
mkhraisha, line 42: Can we make name and value required here?

docs/openapi/components/schemas/common/ExternalIdentifier.yml
```diff
@@ -0,0 +1,49 @@
...
+example: |-
+  {
```
mkhraisha, line 44: Nit: Can you update the example to use leiCode with a correct URI? 
It'll make it easier for future users to adopt

docs/openapi/components/schemas/common/Organization.yml
```diff
@@ -104,22 +90,21 @@ properties:
...
-  taxId:
-    title: Tax Identification Number
-    description: >-
-      The Tax / Fiscal ID of the organization or person, e.g., the TIN in the US
-      or the CIF/NIF in Spain.
-    type: string
-    $linkedData:
-      term: taxId
-      '@id': https://schema.org/taxID
   iataCarrierCode:
```
mkhraisha, line 93: Maybe this moves as well to external identifier?

docs/openapi/components/schemas/common/AgricultureInspectionGeneric.yml
```diff
@@ -293,7 +293,13 @@ example: |-
       "broker": {
         "type": ["Organization"],
         "name": "Cole United",
-        "leiCode": "54321351219389121979"
+        "externalIdentifiers": [
+          {
+            "type": ["ExternalIdentifier"],
+            "name": "LEI Code",
+            "value": "OHVJ56TBUFWT9XWMBEJ9"
```
OR13, line 300: In JSON-LD / RDF... you basically always have name value pairs...

the name is the "property", like 

`identifier` which gets expanded to an IRI like: https://schema.org/identifier

The value, is a value for the property and can have various legal RDF types associated.

https://www.w3.org/TR/json-ld11/#typed-values

Here you are making a new RDF Class "ExternalIdentifier", for a key value pair...

Its sorta not the data modeling convention JSON-LD / RDF experts might expect...

Here is a similar example from schema.org

```json
{
 "@context": "https://schema.org/",
 "@type": "Book",
 "name": "Library linked data in the cloud : OCLC's experiments with new models of resource description",
 "author": "Carol Jean Godby",
 "isbn": "9781627052191",
 "identifier": {
 "@type": "PropertyValue",
   "propertyID": "OCoLC",
   "value":  "889647468"
  },
 "sameAs": "http://www.worldcat.org/oclc/889647468"
}
```

Should we use schema.org instead?

docs/openapi/components/schemas/common/ExternalIdentifier.yml
```diff
@@ -0,0 +1,49 @@
+$linkedData:
+  term: ExternalIdentifier
+  '@id': https://w3id.org/traceability#ExternalIdentifier
```
OR13, line 3: yikes to defining our own RDF class for this... we are killing interop with schema.org for no real benefit.

rhofvendahl: Good catch, thanks!

docs/openapi/components/schemas/common/AgricultureInspectionGeneric.yml
rhofvendahl: Sorry, not entirely following.

I'd be happy to find an alternative to creating an additional schema for such a simple use case, but I'm not clear on how else I could define the `Organization.yml` `externalIdentifiers` prop.

I see that `schema.org/identifier` accepts a `PropertyValue` type which seems like it'd work for our use case, but is there any way to define a property to use an external schema?

For reference, the current `externalIdentifiers` prop on `Organization` that'd need to be modified:
```
  externalIdentifiers:
    title: External Identifiers
    description: External identifiers for this organization.
    items:
      $ref: ./ExternalIdentifier.yml
    $linkedData:
      term: externalIdentifiers
      '@id': https://schema.org/identifier

```

docs/openapi/components/schemas/common/ExternalIdentifier.yml
rhofvendahl: @mkhraisha Done

docs/openapi/components/schemas/common/ExternalIdentifier.yml
rhofvendahl: @mkhraisha I updated the example to use a correctly formatted leiCode, but I'm not sure what the standard is for the URI at which the lei code can be looked up. I put down `http://example.gov/lei/`, will that work?

mkhraisha: yes

docs/openapi/components/schemas/common/AgricultureInspectionGeneric.yml
OR13: `externalIdentifiers` is an RDF Predicate.

`ExternalIdentifier` is an RDF Type.


I question why we are defining either of these instead of using an existing term from schema.org or elsewhere.

brownoxford:
Discussed on call. @OR13 still has pending change requests.

PR #710, opened 2023-03-14
CTPAT schema updates based on feedback

nissimsan:
This PR: 
- Introduces a CTPATMember RDF class because of requirements to allow multiple addresses and SCAC code on the organization
- Restricts the parts of Organization not needed for CTPAT members
- Flattens the CTPAT-Member-Location for easier implementation

PR #714, opened 2023-03-14
CTPAT required attributes

nissimsan:
Remaining required attributes on CTPAT.


nissimsan:
These are trivial extensions to https://github.com/w3c-ccg/traceability-vocab/pull/710 which we covered yesterday. Merging to get them out of the way.

PR #716, opened 2023-03-20
CTPAT string types and separation of IOR

nissimsan:
This: 
- adds string types to enumerated fields
- separates importer of record to a dedicated type

### DIFF
--- a/docs/openapi/components/schemas/common/CBPImporterOfRecord.yml
+++ b/docs/openapi/components/schemas/common/CBPImporterOfRecord.yml
@@ -1,0 +1,46 @@
+$linkedData:
+  term: CBPImporterOfRecord
+  '@id': https://w3id.org/traceability#CBPImporterOfRecord
+title: CBP Importer Of Record
+description: US CBP importer of record identifier and itentifier type (IRS, SSN or CBP).
+type: object
+properties:
+  type:
+    type: array
+    readOnly: true
+    const:
+      - CBPImporterOfRecord
+    default:
+      - CBPImporterOfRecord
+    items:
+      type: string
+      enum:
+        - CBPImporterOfRecord
+  number: 
+    title: importerOfRecordNumber
+    description: Importer Of Record Number
+    type: string
+    $linkedData:
+      term: number
+      '@id': https://w3id.org/traceability#CBPImporterOfRecordNumber
+  identifierType:
+    title: Assigned Identifier Type
+    description: IRS, SSN, or CBP Assigned number.
+    type: string
+    enum:
+      - IRS
+      - SSN
+      - CBP
+    $linkedData:
+      term: identifierType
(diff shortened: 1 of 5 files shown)


BenjaminMoe (approved):

docs/openapi/components/schemas/common/CBPImporterOfRecord.yml
```diff
@@ -0,0 +1,46 @@
+$linkedData:
+  term: CBPImporterOfRecord
+  '@id': https://w3id.org/traceability#CBPImporterOfRecord
+title: CBP Importer Of Record
+description: US CBP importer of record identifier and itentifier type (IRS, SSN or CBP).
```
TallTed, line 5:
```suggestion
description: US CBP importer of record identifier and identifier type (IRS, SSN, or CBP).
```

TallTed (changes requested):

docs/openapi/components/schemas/common/CTPAT.yml
```diff
@@ -64,23 +64,34 @@ properties:
         $linkedData:
           term: iataCarrierCode
           '@id': https://onerecord.iata.org/cargo/Company#airlineCode
```
mkhraisha, line 66: I know this isn't part of this PR but this probably wants to reference ` https://onerecord.iata.org/Carrier#airlineCode` as company#airlinecode isn't a thing

docs/openapi/components/schemas/common/CTPATMember.yml
```diff
@@ -48,23 +48,13 @@ properties:
     $linkedData:
       term: iataCarrierCode
       '@id': https://onerecord.iata.org/cargo/Company#airlineCode
```
mkhraisha, line 50: Similar to above comment

mkhraisha (changes requested):

docs/openapi/components/schemas/common/CTPAT.yml
nissimsan: No, wait, `https://onerecord.iata.org/Carrier#airlineCode` _is_ a thing. The IATA ontology explicitly calls that out: 
image

There's just an in consistency with how the ontology is published. I called that out more than half a year ago on the IATA github: 
https://github.com/IATA-Cargo/ONE-Record/issues/149

I'm open to arguments for going with the double-# URI instead. But I feel pretty confident about sticking with my earlier choice: 
1. The IATA spec explicitly says the IRI is `https://onerecord.iata.org/Carrier#airlineCode` 
2. Double-# URIs are not illegal, but discouraged (second # must be interpreted as a literal), further indicating that's the bug and not IATA's actual intention.
Pls open a separate issue if you think this warrants further discussion; I don't think it does.

mkhraisha (approved):

PR #722, opened 2023-03-27
Fix type error in common/TradeLineItem.yml

BenjaminMoe:
This is fixing a small error in Tradeline item where there is no type (string, number, ect) defined for `countryOfOrigin`.

### DIFF
--- a/docs/openapi/components/schemas/common/TradeLineItem.yml
+++ b/docs/openapi/components/schemas/common/TradeLineItem.yml
@@ -61,7 +61,7 @@ properties:
     type: string
   countryOfOrigin:
     title: Country of Origin
-    description: Identify the 2-digit ISO country code of the country of production.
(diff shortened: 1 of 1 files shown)


docs/openapi/components/schemas/common/TradeLineItem.yml
```diff
@@ -62,6 +62,7 @@ properties:
   countryOfOrigin:
     title: Country of Origin
     description: Identify the 2-digit ISO country code of the country of production.
```
TallTed, line 64: There are no digits (numerals) here...
```suggestion
    description: Identify the 2-character ISO country code of the country of production.
```

brownoxford (approved):

mkhraisha (approved):

nissimsan (approved):

PR #724, opened 2023-03-28
Issuer Organization narrowing

nissimsan:
This PR introduces a limited subset of Organization intended to be generically applicable for vc.issuer.

I only updated CTPAT and CertificateOfOrigin for now, but I would consider this (or another dedicated narrowing) anywhere where we currently use: 

```
  issuer:
    $ref: ../common/Organization.yml
```

### DIFF
--- a/docs/openapi/components/schemas/snippets/IssuerOrganization.yml
+++ b/docs/openapi/components/schemas/snippets/IssuerOrganization.yml
@@ -1,5 +1,5 @@
-title: Intent to Import Organization
-description: An organization such as a corporation, firm, club, etc.
+title: Issuer Organization
+description: Organization issuing a Verifiable Credential.
 type: object
 properties:
   type:
@@ -61,9 +61,6 @@ properties:
             title: Name
             description: The name of the entity in text.
             type: string
-            $linkedData:
-              term: name
-              '@id': https://schema.org/name
           streetAddress:
             title: Street Address
             description: >-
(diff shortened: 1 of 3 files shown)


BenjaminMoe (approved):

docs/openapi/components/schemas/snippets/IssuerOrganization.yml
```diff
@@ -0,0 +1,112 @@
...
+    title: Name
+    description: Name of the organization.
+    type: string
+  description:
+    title: Description
+    description: Description of the company.
+    type: string
+  location:
+    title: Location
+    description: The location of, for example, where an event is happening, where an organization is located, or where an action takes place.
```
TallTed, line 30:
```suggestion
    description: The location where, for example, an event is happening, an organization is located, or an action takes place.
```

docs/openapi/components/schemas/snippets/IssuerOrganization.yml
```diff
@@ -0,0 +1,112 @@
...
+          name:
+            title: Name
+            description: The name of the entity in text.
+            type: string
+          streetAddress:
+            title: Street Address
+            description: >-
+              The street address expressed as free form text. The street address is
+              printed on paper as the first lines below the name. For example, the name
+              of the street and the number in the street or the name of a building.
```
TallTed, line 69:
```suggestion
              of the street and the number in the street, or the name of a building.
```

TallTed (changes requested):

mkhraisha (approved):

docs/openapi/components/schemas/snippets/IssuerOrganization.yml
```diff
@@ -0,0 +1,112 @@
+title: Issuer Organization
+description: Organization issuing a Verifiable Credential.
+type: object
+properties:
+  type:
+    type: array
+    readOnly: true
```
brownoxford, line 7: Doesn't `readOnly` in this context imply that this value must not be present when submitting a request that contains this type of object in the request body, but it will be present in responses containing this type of object?

nissimsan: Hey Chris, no the contrary, this is to tell a form renderer that you got not choice but to include it. 
https://jsonforms.io/docs/readonly/

We do this everywhere for `type` and `@context`.

Issue #536, opened 2022-08-18
Align to the production UN Vocab


nissimsan:
The UN CEFACT LD vocab should be bumped to version 1.0, expected during fall of 2022.

The main updates which will be required to be update on our side include: 
- Switch away from `/#` (see [286](https://github.com/w3c-ccg/traceability-vocab/issues/268) and [531](https://github.com/w3c-ccg/traceability-vocab/issues/531)).
- Code lists switching from description to code (https://github.com/uncefact/spec-jsonld/issues/24).

I advise that we don't start this work until the UN vocab is published.

nissimsan:
Current draft version: https://service.unece.org/trade/uncefact/vocabulary/uncefact/
Repo: https://github.com/uncefact/spec-jsonld

nissimsan:
@VladimirAlexiev, @brownoxford, FYI ^

BenjaminMoe:
@nissimsan any updates?

nissimsan:
Yes, it's basically done. we're currently waiting for the `vocabulary.uncefact.org` DNS to propagate. It takes oddly long, something must have gone wrong. 

This is what v1 will look like, though: dmvc7xzscpizo.cloudfront.net

nissimsan:
The much improved production vocabulary.uncefact.org is live now. We should switch our pointers from the draft URIs to this. 
For example:
https://service.unece.org/trade/uncefact/vocabulary/uncefact/#consigneeParty
should be changed to 
https://vocabulary.uncefact.org/consigneeParty

nissimsan:
We did this!

mgh128:
@nissimsan  - re "We did this!", I'm wondering what exactly we did.

I've just been looking at https://vocabulary.uncefact.org/UnitMeasureCode

Hyperlinks such as https://vocabulary.uncefact.org/UnitMeasureCode#KGM go nowhere / provide no further details and I don't see any details about conversion factors when I check the source code for the page.  

I also tried reloading the page after setting the HTTP header Accept:  application/ld+json  but that just produced a 404 error page with this rather unfriendly message:

404 Not Found
Code: NoSuchKey
Message: The specified key does not exist.
Key: UnitMeasureCode.jsonld
RequestId: KQDKKWY3RJQ844FQ
HostId: CFF2nAkAhmrgaZgiqdJx8hrd7djYmRIsn9XamQ/2YY3MrwyjNkyG1tt43OxWM+p5kbp9hyRNIiY=

[...]

PR #726, opened 2023-03-29
UN/CEFACT production term update

nissimsan:
As [announced a while back](https://github.com/w3c-ccg/traceability-vocab/issues/536) the UN/CEFACT vocab has been promoted from previous draft state to production. [The project](https://uncefact.unece.org/display/uncefactpublic/JSON-LD+Web+Vocabulary) is not yet formally closed, but development efforts are considered to be completed. 

https://vocabulary.uncefact.org/

The main difference from the draft UN vocab is a much nicer DNS and path. But there are also some NDR updates, so a dirty search-replace wasn't enough, I made sure each URI resolves and makes business sense. 

This will obviously cause breaking proofs across the board, changing the terms behind our context. There is never a good time to do this kind of thing.


mprorock:
better now than later - really appreciate the updates on this

BenjaminMoe (approved):

PR #727, opened 2023-03-29
Reduce Instances of Organization

BenjaminMoe:
Any where where `Organization` is referenced directly makes the schema massive with respect for attributes which are not required to be there. 

This PR reduces the direct use of `Organization` in issuer and `TradeLineItem`.

### DIFF
--- a/docs/openapi/components/schemas/common/TradeLineItem.yml
+++ b/docs/openapi/components/schemas/common/TradeLineItem.yml
@@ -51,7 +51,200 @@ properties:
       '@id': https://service.unece.org/trade/uncefact/vocabulary/uncefact/#packageQuantity
   product:
     title: Product
-    $ref: ./Product.yml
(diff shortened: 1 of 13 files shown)


docs/openapi/components/schemas/common/TradeLineItem.yml
```diff
@@ -51,7 +51,200 @@ properties:
...
+            description: Organization's contact phone number.
+            type: string
+          logo:
+            type: string
+            format: uri
+            description: Logo of the entity.
+          taxId:
+            title: Tax Identification Number
+            description: >-
+              The Tax / Fiscal ID of the organization or person, e.g., the TIN in the US
```
TallTed, line 125:
```suggestion
              The Tax or Fiscal ID of the organization or person, e.g., the TIN in the US
```

PR #736, opened 2023-04-12
Add v2 test suite

OR13:
This PR adds and experimental test suite for VCDM v2.

Current approach transforms the published schemas, but in the future, when VCDM v2 is a TR, this transformation should not be necessary.


packages/traceability-tests/examples/v2.json
```diff
@@ -0,0 +1,10509 @@
...
+      "verified": {
+        "protectedHeader": {
+          "iss": "did:jwk:eyJrdHkiOiJPS1AiLCJjcnYiOiJFZDI1NTE5IiwieCI6InpYMnpvR21XMzZUVExfa3czZy1LRlZqaDVJb2FEY29TYzA4aklENHVHcnMifQ",
```
OR13, line 13: @mprorock note the header params.

packages/traceability-tests/examples/v2.json
```diff
@@ -0,0 +1,10509 @@
...
+          "kid": "#0",
+          "alg": "EdDSA",
+          "cty": "vc+ld+json"
```
OR13, line 16: Note the v2 content type.

packages/traceability-tests/examples/v2.json
```diff
@@ -0,0 +1,10509 @@
...
+              ],
+              "streetAddress": "19 Knox St",
+              "addressLocality": "Toronto",
+              "addressRegion": "ON",
+              "addressCountry": "Canada",
+              "postalCode": "M3B 1A2"
+            }
+          },
+          "credentialSchema": {
+            "id": "https://w3id.org/traceability/openapi/components/schemas/credentials/BankAccountCredential.yml",
```
OR13, line 59: Note the self referencing credentialSchema... this is checked... AFTER ... signature verification in the tests.

packages/traceability-tests/examples/v2.json
```diff
@@ -0,0 +1,10509 @@
...
+      "verified": {
+        "protectedHeader": {
+          "iss": "did:jwk:eyJrdHkiOiJPS1AiLCJjcnYiOiJFZDI1NTE5IiwieCI6InpYMnpvR21XMzZUVExfa3czZy1LRlZqaDVJb2FEY29TYzA4aklENHVHcnMifQ",
+          "kid": "#0",
+          "alg": "EdDSA",
+          "cty": "vc+ld+json"
+        },
+        "payload": {
+          "@context": [
+            "https://www.w3.org/ns/credentials/v2",
```
OR13, line 20: v2 context

OR13:
This test suite won't pass, because not all schemas support `credentialSchema` property... given how fundamental to our work this is, it seems like we should correct this.

mprorock:
> This test suite won't pass, because not all schemas support `credentialSchema` property... given how fundamental to our work this is, it seems like we should correct this.

@rhofvendahl might want to take a pass on this to make sure we have proper `credentialSchema` support in place

mprorock (approved):
@brownoxford I like the approach taken here, particularly with did:jwk instead of did:key for testing - this will cause us to need to update a few things as we shift more to this model

packages/traceability-tests/tests/jsonld.test.js
```diff
@@ -0,0 +1,33 @@
...
+  await kyPromise;
+});
+
+it('canonize', async () => {
+  expect(
+    await canonize(
+      {
+        '@context': { '@vocab': 'https://www.w3.org/2018/credentials#' },
+        type: ['VerifiableCredential'],
+        issuer: 'https://example.edu/issuers/565049',
```
mprorock, line 16:
```suggestion
        issuer: 'https://example.org/issuers/565049',
```

packages/traceability-tests/tests/jsonld.test.js
```diff
@@ -0,0 +1,33 @@
...
+      },
+      {
+        format: 'application/n-quads',
+      }
+    )
+  )
+    .toBe(`_:c14n0 <https://www.w3.org/2018/credentials#id> "did:example:ebfeb1f712ebc6f1c276e12ec21" .
+_:c14n1 <https://www.w3.org/2018/credentials#credentialSubject> _:c14n0 .
+_:c14n1 <https://www.w3.org/2018/credentials#issuanceDate> "2010-01-01T19:23:24Z" .
+_:c14n1 <https://www.w3.org/2018/credentials#issuer> "https://example.edu/issuers/565049" .
```
mprorock, line 30:
```suggestion
_:c14n1 <https://www.w3.org/2018/credentials#issuer> "https://example.org/issuers/565049" .
```

packages/traceability-tests/tests/privateKeyJwk.js
```diff
@@ -0,0 +1,8 @@
+const privateKeyJwk = {
+  kty: 'OKP',
+  crv: 'Ed25519',
+  x: 'zX2zoGmW36TTL_kw3g-KFVjh5IoaDcoSc08jID4uGrs',
+  d: '95hNcacuqtWQP70MDYrL6DRUJ8iGg-oq13I0KM5Etw8',
+};
+
+module.exports = privateKeyJwk;
```
mprorock, line 8: This works, but perhaps p256?

OR13: I can change that part, it just makes the examples longer.

nissimsan (approved):

.github/workflows/v2.test.yml
```diff
@@ -0,0 +1,19 @@
...
+on: [push]
+
+jobs:
+  ci:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Begin CI...
+        uses: actions/checkout@v2
+      - name: Use Node 16
+        uses: actions/setup-node@v1
```
mkhraisha, line 12: use node v3? and checkout v3? also might be a good idea to use cache with it to speed it up.

packages/traceability-tests/tests/issue-verify.test.js
```diff
@@ -0,0 +1,82 @@
+const jose = require('jose');
+const fs = require('fs');
+const { v4: uuidv4 } = require('uuid');
+// pending merge of https://github.com/w3c-ccg/traceability-vocab/pull/735
```
mkhraisha, line 4: putting a comment here to make sure this gets changed after pr 735 is merged but before this pr is.

mkhraisha (approved):

OR13:
Schema validation requires: https://github.com/w3c-ccg/traceability-vocab/pull/747

.github/workflows/v2.test.yml
OR13: Do you have a snippet you would prefer I use here?

OR13:
I should be unblocked, but we might find out more errors in schemas after we are unblocked.

.github/workflows/v2.test.yml
brownoxford: See any of the workflows in trace-interop:
```
...
    steps:
      # Check out repo, set up node, and install dependencies.
      # @see https://github.com/actions/setup-node#usage
      - uses: actions/checkout@v3
      - uses: actions/setup-node@v3
        with:
          node-version: 16.15.1
          cache: 'npm'
      - run: npm ci
...
```

OR13:
Several related issues have been uncovered here...

- https://github.com/w3c-ccg/traceability-vocab/issues/786
- https://github.com/w3c-ccg/traceability-vocab/issues/787

We will probably want to refactor the schema package tests next, so we can start issuing both v1 and v2 credentials as part of CI tests.

PR #748, opened 2023-04-19
Updates to MTR

OR13:


docs/openapi/components/schemas/common/ChemicalComposition.yml
```diff
@@ -0,0 +1,153 @@
+$linkedData:
+  term: ChemicalComposition
+  '@id': https://w3id.org/traceability#ChemicalComposition
+title: Chemical Composition
+description: Measured chemical composition; all values in W%.
```
mkhraisha, line 5: Im guessing W% is a typo?

mkhraisha (commented):
We have: https://github.com/w3c-ccg/traceability-vocab/blob/main/docs/openapi/components/schemas/common/ChemicalProperty.yml 
for chemical Properties, and now we're adding chemical composition, which seems to be steel specific. 
We should keep in mind that other items may need a chemical composition and that it would require them to add their own terms to this. 
Is there a way we can use Chemical Property to make it more extensible?

docs/openapi/components/schemas/common/ChemicalComposition.yml
TallTed: I think it stands for "[Weight percentage](https://www.google.com/search?q=%22W%25%22+weight)"... which would probably be better written out, here.
```suggestion
description: Measured chemical composition; all values in weight percentage (W%). 
```

nissimsan:
Yes, @TallTed , weight percentage - thank you for clarifying!

BenjaminMoe (approved):

nissimsan (approved):

BenjaminMoe (changes requested):
Should be inlined.

mkhraisha (approved):

BenjaminMoe (approved):

mprorock (approved):

PR #753, opened 2023-04-26
Inspection min length

nissimsan:
Adding minLength of 1, an MTR without inspection is useless.


nissimsan:
Merging this out of band in the process of MTR refinement.

PR #758, opened 2023-04-28
MTR shipto optionality

nissimsan:
Optional elements should not contain mandatory child elements.

### DIFF
--- a/docs/openapi/components/schemas/credentials/MillTestReportCredential.yml
+++ b/docs/openapi/components/schemas/credentials/MillTestReportCredential.yml
@@ -353,20 +353,8 @@ properties:
                     description: Text specifying the postal code for an address.
                     type: string
                 additionalProperties: false
-                required:
-                  - type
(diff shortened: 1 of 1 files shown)


mkhraisha (approved):

OR13 (approved):

TallTed:
> Optional elements should not contain mandatory child elements.

Would these not be conditionally mandatory?  That is, if an optional element is present, it must contain specified child elements. If the optional element is not present, those child elements are not required because their parent element is not present...

mkhraisha:
Can we add the "optional mandatory fields" into the description of the PR and merge it in out of band.

brownoxford:
Would it not be better to leave these in and add comments indicating that they should be commented out if you use one of the broken tools?

TallTed:
Most (probably all) instances of "Conditionally mandatory if..." would be better just "Mandatory if..."

PR #765, opened 2023-05-03
fix: grade min items

nissimsan:
Super tiny.


nissimsan:
Merging this out of bound, it's "our" schema and so small its not even worth mentioning.

PR #766, opened 2023-05-04
MTR minor tweaks

nissimsan:
Based on review of more real life samples, this: 
1. Adds Lead and Carbon Equivalence chemical analysis
2. Moves product desc from inspection array to root 
3. Fixes a bad measured value description
4. Makes steel grade a string instead of array

### DIFF
--- a/docs/openapi/components/schemas/credentials/MillTestReportCredential.yml
+++ b/docs/openapi/components/schemas/credentials/MillTestReportCredential.yml
@@ -220,7 +220,6 @@ properties:
                   streetAddress:
                     title: Street Address
                     description: >-
-                      The street address expressed as free form text. The street address is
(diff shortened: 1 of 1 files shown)


docs/openapi/components/schemas/credentials/MillTestReportCredential.yml
```diff
@@ -220,7 +220,6 @@ properties:
                   streetAddress:
                     title: Street Address
                     description: >-
-                      The street address expressed as free form text. The street address is
                       printed on paper as the first lines below the name. For example, the name
```
TallTed, line 223:
```suggestion
                      Typically printed on paper as the first lines below the name. For example, the name
```

BenjaminMoe (approved):

OR13 (approved):

PR #772, opened 2023-05-11
adds deliverySchedule and Statement credentials

mkhraisha:


nissimsan (approved):

docs/openapi/components/schemas/credentials/DeliveryScheduleCredential.yml
```diff
@@ -0,0 +1,256 @@
...
+  issuanceDate:
+    type: string
+  issuer:
+    $ref: ../common/Organization.yml
+  credentialSchema:
+    type: object
+    properties:
+      id:
+        title: Id
+        description: The url of the schema file to validate the shape of the json object
```
TallTed, line 54:
```suggestion
        description: The URL of the schema file to use to validate the shape of the JSON object
```

docs/openapi/components/schemas/credentials/DeliveryScheduleCredential.yml
```diff
@@ -0,0 +1,189 @@
...
+        },
+        "address": {
+          "type": [
+            "PostalAddress"
+          ],
+          "streetAddress": "",
+          "addressLocality": "Hardisty",
+          "addressRegion": "AB",
+          "postalCode": "",
+          "addressCountry": "CAN"
```
nissimsan, line 131: We are recommending ourselves to use 2-char ISO country codes. Let's align to our own recommendation.

nissimsan (changes requested):

docs/openapi/components/schemas/common/DeliverySchedule.yml
```diff
@@ -215,76 +215,8 @@ example: |-
...
-        "addressLocality":"Leannonland",
-        "addressRegion":"Georgia",
-        "postalCode":"48950-9803",
-        "addressCountry":"French Polynesia"
-      },
-      "unLocode":"DKCPH"
-    },
+    "portOfEntry": "3501",
+    "portOfDestination": "3901",
+    "portOfArrival": "4105",
```
OR13, line 220: where do we find these codes?

mkhraisha:
Add a description/ JSON-LD term update to the delivery schedule codes to make it specific which coding we're using

OR13 (approved):

OR13 (approved):

docs/openapi/components/schemas/common/DeliverySchedule.yml
OR13: Given these are CBP port codes, this credential should probably be titled CBPDeliverySchedule... since nobody else can use it.

https://www.cbp.gov/sites/default/files/assets/documents/2020-Jan/ACE%20Appendix%20E%20Schedule%20D%20January%202020%20%281%29.pdf

TallTed: Or just maybe `USACBPDeliverySchedule`, since there's no guarantee other nations will never have their own CBP and/or `CBPDeliverySchedule`

PR #774, opened 2023-05-12
Add section on catair mapping

OR13:
This PR adds a simple way to include additional html next to credentials with a specific tag.


nissimsan:
@OR13, the attribute-level bottom-up approach hinted at here seems somewhat unrealistic. 

I suggest we rather start from the top. This document is a great starting point: 
https://www.cbp.gov/sites/default/files/assets/documents/2017-Oct/ACE_CATAIR_TableOfContents_June_2004.pdf

I would first map at credential-level to the CATAIR ToC. Then later on, we figure out how to break it further down appropriately.

docs/sections/catair-mapping.html
```diff
@@ -0,0 +1,50 @@
...
+    and how certain data elements might be new, 
+    and will have no mapping to CATAIR.
+  </p>
+
+  <table class="simple" >
+    <thead>
+      <tr>
+        <td>JSON Pointer</td>
+        <td>Value</td>
+        <td>CATAIR</td>
```
nissimsan, line 28: Too detailed. Let's break this down before getting into attribute-level.

nissimsan (changes requested):

docs/sections/catair-mapping.html
OR13: The original request was "per attribute", for "each credential".

mkhraisha (changes requested):
My understanding is the requirement coming in is what are the fields that are NOT in CATAIR

docs/sections/catair-mapping.html
nissimsan: I see, I misunderstood. Then should we move this to the Credential section instead. That would make it clearer.

nissimsan (approved):

mkhraisha (changes requested):
We should be able to inline HTML next to every credential instead of in its own section, for MTR this should be in the MTR credential area.

OR13:
We discussed on the call, I will revise.

docs/openapi/components/schemas/credentials/MillTestReportCredential.yml
```diff
@@ -4,6 +4,7 @@ $linkedData:
 title: Mill Test Report Credential
 tags:
   - Steel
+  - TechDemo2023
```
OR13, line 7: Add a tag like this.

docs/sections/catair/MillTestReportCredential.html
```diff
@@ -0,0 +1,26 @@
+<table class="simple" >
```
OR13, line 1: Add a file like this.

packages/traceability-schemas/scripts/schemas-to-vocab.js
```diff
@@ -76,15 +80,25 @@ const buildClass = (schema) => {
 
   const depList = dependencies.length
     ? `<b>Dependencies</b><ul>${dependencies
-        .map((key) => `<li><a href='#${key}'>${key}</a></li>`)
+        .map((key) => `<li>${key}</li>`)
```
OR13, line 83: this was causing all the respec errors... snippets had no valid anchor definition.

OR13:
@mkhraisha please re-review.

Screen Shot 2023-05-16 at 6 17 27 PM

docs/openapi/components/schemas/credentials/MillTestReportCredential.yml
mkhraisha: This should probably be a catair tag not a techdemo2023 tag

mkhraisha (commented):
Probably a good idea to update the README with instructions on how to do this

OR13:
@mkhraisha the entire script process to generate the spec is pretty gross... I would really like to refactor the whole thing, but that would be disruptive, here so I am holding off.

nissimsan:
Neither @OR13  nor @mkhraisha  are on the call - leaving this PR for next week.

mkhraisha (approved):
Changes look great thank you!

PR #779, opened 2023-05-17
ISF sector

nissimsan:
I don't believe oil or gas is ever containerized.


mprorock (approved):
Looks like a correct approach - [bulk cargo](https://www.ecfr.gov/current/title-19/chapter-I/part-149#p-149.4(a)) is exempted and oil/gas falls under this:

BenjaminMoe (approved):

PR #781, opened 2023-05-17
Entry number credential

nissimsan:
This is a proposal to satisfy the "entry number only" requirement with a credential which is more business relatable and fit for purpose. 

It also has a stronger LD term, which I also aligned with 3461 and 7501 too.


mprorock (approved):
This is a good simplification and the pointer to cefact is good

docs/openapi/components/schemas/common/CBPEntrySummary.yml
```diff
@@ -22,7 +22,7 @@ properties:
     type: string
     $linkedData:
       term: entryNumber
-      '@id': https://schema.org/identifier
+      '@id': https://vocabulary.uncefact.org/ReferenceCodeList#AQM
```
mkhraisha, line 25: This is great

mkhraisha (changes requested):
Few issues with this idea.

1. By removing Thing we have no way of allowing an identifier that exists on an outside system to be connected to a workflow in a generic way, we are committing to adding many of these smaller credentials. 
2. This removes the ability to add a free trade zone number.

nissimsan:
@mkhraisha, how about I leave ThingCredential untouched and just add EntryNumberCredential (instead of overwriting)?

mkhraisha:
We could do that or we can all agree on this direction and then I'll add a FTZ no credential.


Regardless I think we may need the thing credential.

docs/openapi/components/schemas/common/EntryNumber.yml
```diff
@@ -0,0 +1,36 @@
...
+      - EntryNumber
+    default:
+      - EntryNumber
+    items:
+      type: string
+      enum:
+        - EntryNumber
+      entryNumber:
+        title: Entry Number
+        description: The 11 digit alphanumeric code. The entry number is comprised of the three-digit filer code, followed by the seven-digit entry number, and completed with the one-digit check digit. The Entry Filer Code represents the three-character alphanumeric filer code assigned to the filer or importer by CBP. The Entry Number represents the seven-digit number assigned by the filer. The number may be assigned in any manner convenient, provided that the same number is not assigned to more than one CBP Form 3461. Leading zeros must be shown. The check digit is computed on the previous 10 characters. The formula for calculating the check digit can be found in Appendix 1.
```
TallTed, line 21:
```suggestion
        description: The 11 digit alphanumeric code. The entry number is comprised of the three-digit filer code, followed by the seven-digit entry number, and completed with the one-digit check digit. The Entry Filer Code represents the three-character alphanumeric filer code assigned to the filer or importer by CBP. The Entry Number represents the seven-digit number assigned by the filer. The number may be assigned in any convenient manner, provided that the same number is not assigned to more than one CBP Form 3461. Leading zeros must be shown. The check digit is computed on the previous 10 characters. The formula for calculating the check digit can be found in Appendix 1.
```

docs/openapi/components/schemas/credentials/EntryNumberCredential.yml
```diff
@@ -1,14 +1,14 @@
...
+  '@id': https://w3id.org/traceability#EntryNumberCredential
+title: Entry Number Credential
 tags:
   - Steel
   - eCommerce
   - Agriculture
   - Oil and Gas
   - Other
-description: A credential that contains a thing.
+description: A credential which associates an customs entry number to a workflow.
```
TallTed, line 11:
```suggestion
description: A credential which associates a customs entry number to a workflow.
```

nissimsan:
@mkhraisha, ThingCred is added back. 

I don't know the detailed requirements going into a TFZ credential, so leaving that to you to model.  https://vocabulary.uncefact.org/LocationFunctionCodeList#273 seems like an applicable term.

mkhraisha (approved):

BenjaminMoe (approved):

Issue #640, opened 2022-11-29
For mock numbers "555" should be prefix, not area code


rhofvendahl:
AFAIK good practice for mock phone numbers is to use 555 as the prefix (the three digits after area code), along the lines of "(xxx) 555-xxxx"). [Here's](https://en.wikipedia.org/wiki/555_(telephone_number)) a wiki article on use of the 555 prefix, and [another](https://en.wikipedia.org/wiki/Fictitious_telephone_number) on mock phone numbers generally.

Currently most of the mock numbers in our schema examples either use 555 as the area code, or not at all. I propose we update all numbers to use the format XXX-555-XXXX.

I'd be willing to implement this, if there are no objections.

rhofvendahl:
Also: I'd be interested in hearing opinions on `+1-xxx-xxx-xxxx` format vs just `xxx-xxx-xxxx`. Currently it's a mix, is that what we want?

TallTed:
`+1` is the country code for USA, Canada, and various smaller entities that also participate in the [NANPA](https://en.wikipedia.org/wiki/North_American_Numbering_Plan).  It is appropriate to use this for all such (fake or real) numbers.

Any fake numbers we're putting in other countries should get the relevant (real) country code, e.g., `+44` for the UK.

Omitting the country code may lead to various mischief including (intended or not) DDOS attacks by software being tested (or deployed!) with some connection to the fake data we're providing... because by omitting (for instance) the `+1`, the first dialed digits may be the country code for another country....

All of which is to say — I'd convert from the mix to always include the `+1` on NANPA numbers.

OR13:
I think we did this, close it

nissimsan:
We did this.

PR #783, opened 2023-05-17
Format mock numbers

rhofvendahl:
This PR applies appropriate formatting for mock numbers across all schemas, as discussed in https://github.com/w3c-ccg/traceability-vocab/issues/640.


TallTed (approved):
I've not looked at every number, but those I have looked at appear to do the right thing.

OR13 (approved):

docs/openapi/components/schemas/common/DCSAShippingInstruction.yml
```diff
@@ -208,7 +208,7 @@ example: |-
         }
       },
       "email": "xxinau-sales@example.org",
-      "phoneNumber": "+86-555-865-8495"
+      "phoneNumber": "+86-074-555-4172"
```
mkhraisha, line 211: Are we keeping +89? 
I thought we're moving to all +1 with 555 area codes as that is basically guaranteed to be fake.

mkhraisha (approved):

docs/openapi/components/schemas/common/DCSAShippingInstruction.yml
rhofvendahl: @mkhraisha That's a good question. Changing everything to +1 with 555 area codes would be US-centric but safe. Ideally we could include fake numbers for a variety of country schemes, but this gets tricky as different countries have [different conventions](https://en.wikipedia.org/wiki/Fictitious_telephone_number) for fake numbers - which'd be annoying to implement and keep track of I think.

In this PR I've maintained the format & country codes of non-US numbers, but swapped out the equivalent of area code for 555. My thinking was that this'd be easy to keep track of (e.g. checking whether a fake number is formatted correctly) & make it clear that these numbers should be treated as fake, but depending on the country & its conventions this isn't 100% safe as far as ensuring fake numbers.

I'm open to changing this though, if you think it'd be better to just swap things for +1 with 555 numbers or else take some other approach.

BenjaminMoe (approved):

nissimsan (approved):

nissimsan:
@rhofvendahl , please go ahead and merge when the conflict is resolved.

PR #785, opened 2023-05-19
Add workflows to respec doc

BenjaminMoe:
This pull request creates a workflow directory and adds a simple example for testing. 

image


docs/sections/workflows.html
```diff
@@ -0,0 +1,34 @@
+
+    <section>
+      <h2>Workflows</h2>
+      
+      <h3>Business Card Workflow</h3>
+      <image style='border: 1px solid #9a9a9a; border-radius: 3px; background-color: #2a303c;' src='resources/wf_businesscard.svg'/>
+      <p>This is a simple as possible workflow that describes an interaction between  two individuals. Alice and Bob are two individuals who support sending  presentations using the Traceability API. Alice issues a credential and  sends it to Bob in a Traceable Presentation.</p>
+      <b>Credentials Used:</b>
+      <ol>
+        <li><a href="undefined">Verifiable Business Card</a></li>
```
OR13, line 10: why undefined?

packages/traceability-schemas/scripts/schemas-to-vocab.js
```diff
@@ -129,9 +121,49 @@ const separateSchemas = (schemaList) => {
   return { credentials, vocab };
 };
 
+const buildWorkflowSection = () => {
+  const images = {
+    "Business Card Workflow": "resources/wf_businesscard.svg"
```
OR13, line 126: why use mermaid, if you we don't render it?

https://mermaid.js.org/config/usage.html

^ lets do this, for the `mermaid` block... and maybe add an example that uses an image, so we cover both cases.

nissimsan (approved):

packages/traceability-schemas/scripts/schemas-to-vocab.js
nissimsan, line 126: Having to add the image reference here seems wrong. The workflow yml should be self-contained. 

How about adding the SVG as an element alongside mermaid. 

```
mermaid: >-
  sequenceDiagram
      participant A as Alice
      ...
svg: >-
  <svg aria-roledescription="sequence"
  ...
    
```

nissimsan (approved):

docs/sections/workflows.html
BenjaminMoe: i changed from `url` to `id`, I can update this.

brownoxford (approved):

PR #788, opened 2023-05-22
US CBP Import Workflows

nissimsan:
This PR builds on @BenjaminMoe's work, adding workflows related different aspects of US CBP import regulation.


BenjaminMoe:
The workflow definitions are loaded in dynamically after the page init, so i had to add a function to check for the diagrams, render them and then append the images. 

image

BenjaminMoe (approved):

brownoxford (approved):

PR #791, opened 2023-05-26
Importer number alignment

nissimsan:
This aligns 7501 with how IOR is defined on 3461 and CTPAT.

### DIFF
--- a/docs/openapi/components/schemas/common/CBPEntrySummary.yml
+++ b/docs/openapi/components/schemas/common/CBPEntrySummary.yml
@@ -199,10 +199,10 @@ properties:
   importerNumber:
     title: Importer Identifier
     description: Record the IRS EIN, SSN, or CBP assigned number of the importer of record. Proper format is listed under the instructions for Consignee Number.
-    type: string
+    $ref: ./CBPImporterOfRecord.yml
     $linkedData:
       term: importerNumber
-      '@id': https://schema.org/identifier
+      '@id': https://w3id.org/traceability#importerOfRecord
   referenceNumber:
     title: Reference Number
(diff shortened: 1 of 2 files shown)


docs/openapi/components/schemas/common/CBPEntrySummary.yml
```diff
@@ -199,10 +199,10 @@ properties:
   importerNumber:
     title: Importer Identifier
     description: Record the IRS EIN, SSN, or CBP assigned number of the importer of record. Proper format is listed under the instructions for Consignee Number.
-    type: string
+    $ref: ./CBPImporterOfRecord.yml
```
mkhraisha, line 202: Why would we do this? 
This is a CBP entry summary when would you not have the importerNumber be the CBPImporter of Record?

OR13 (approved):

docs/openapi/components/schemas/common/CBPEntrySummary.yml
```diff
@@ -199,10 +199,10 @@ properties:
...
     $linkedData:
       term: importerNumber
-      '@id': https://schema.org/identifier
+      '@id': https://w3id.org/traceability#importerOfRecord
```
OR13, line 205: sad that we define this

OR13:
I'm fine merging this outside of call time.

nissimsan:
@nissimsan , merge when sorted out with @mkhraisha

docs/openapi/components/schemas/common/CBPEntrySummary.yml
nissimsan: Good point. 

How about `Declarant's Customs identity number`: https://vocabulary.uncefact.org/ReferenceCodeList#ABP

docs/openapi/components/schemas/common/CBPEntrySummary.yml
nissimsan: Indeed you have one of the applicable importer identifiers: IRS EIN, SSN, or CBP assigned. The class is to allow indication which one it is. We use it elsewhere, this is to make 7501 consistent.

PR #798, opened 2023-06-05
Schema cleanup

nissimsan:
This removes a bunch of schema elements which are not used: 
- name
- description
- relatedLinks


BenjaminMoe (approved):

nissimsan:
Awaiting Mesur to add an approval on this. Then agreement on the call to merge out of band.

brownoxford:
@rhofvendahl please take point on this research

rhofvendahl:
@nissimsan There are a number of ag related schemas which include include `name`, `description`, and `relatedLink` as fields, but in most cases only `name` is used in the example.

@mkhraisha you're saying that it'd be fine to leave these `name` values in the example even if we remove that yaml field? & this holds true even with `additionalProperties` set to `false`?

I'm not sure I understand the pattern we're following if that's the case. Are we only including yaml fields that are both in the spec and in use in examples?

nissimsan:
@rhofvendahl, that should not be the case. Do you mean issuer's name? That's something else. 

If there are indeed a few credential name I've missed, please point me to which ones they are and I'll clean it out.

rhofvendahl:
@nissimsan If I'm understanding the intent here it looks as though most ag-related schemas plus a few others have the name & description properties, including:
`CTPATCertificate`
`DeliveryScheduleCredential`
`DeliveryStatementCredential`
`FoodDefenseInspectionCredential`
`FoodGradeInspectionCredential`
`FSMACreatingCTECredential`
`FSMAFirstReceiverDataCredential`
`FSMAGrowingCTECredential`
...

(might make more sense to just search for it actually)

Like I said a number of these do have values for `name` & `relatedLink` in the example - I'm not sure how to handle that

nissimsan:
@rhofvendahl, correct. 

All the latter ones are because I didn't want to touch the Oil+Gas and Agri schemas. I definitely encourage that you have similar considerations on "your" schemas and purge unused fields. 

The CTPAT still has name and descr because it makes business sense for it to include them, providing holders and verifiers a human readable context of the certificate.

rhofvendahl:
@nissimsan Ah, that makes sense. No objections then.

PR #806, opened 2023-06-09
Merge invoices into single credential

BenjaminMoe:
- Merges `ProFormaInoviceCredential` and `CommercialInvoiceCredential` into `InvoiceCredential`
- Adds attribute `isProforma` to indicate if an invoice is in the proforma state


docs/openapi/components/schemas/credentials/InvoiceCredential.yml
```diff
@@ -1,7 +1,7 @@
 $linkedData:
-  term: CommercialInvoiceCredential
-  '@id': https://w3id.org/traceability#CommercialInvoiceCredential
-title: Commercial Invoice Credential
+  term: InvoiceCredential
```
nissimsan, line 2: @BenjaminMoe, I appreciate that you want to avoid definition redundancy - similar to how they used to share `$ref:Invoice.yml`. But from a "recognizable by the industry" point of view, this seems like a step backwards. People are used to forms with a "Commercial Invoice" header or "Proforma Invoice" header. I suggest we accept redundancy and keep the individual types of credentals.

nissimsan (changes requested):

mprorock (approved):
This is a good step in the right direction.  Obviously business logic can display, export, render, etc however desired.

docs/openapi/components/schemas/credentials/InvoiceCredential.yml
BenjaminMoe: PR moved to https://github.com/w3c-ccg/traceability-vocab/pull/808

PR #814, opened 2023-06-15
Adding US CBP 4 digit port code

nissimsan:
US port codes, required on CBP forms.


mprorock:
I totally get this for entry forms, but un locodes and NGA identifiers are pretty common as well - any thoughts as to ensuring we keep broader port info available in the vocab?

A situation we have been tracking often, is cases where there is a known inbound port identified by US codes, but multiple stops along the way we need to track that are not necessary "on the map".

mprorock (approved):

BenjaminMoe (approved):

OR13 (approved):

PR #815, opened 2023-06-15
Update Requirements

OR13:


docs/index.html
```diff
@@ -1131,22 +1131,21 @@ <h3>Extensions to Standard</h3>
 
         <section>
           <h4>Verifiable Credential</h4>
-
           <p>The object MUST conform to the basic requirements section of the <a href="https://www.w3.org/TR/vc-data-model/#basic-concepts">W3C VC Data Model</a>.</p>
-
-          <p>The <code>@context</code> MUST contain <code>"https://w3id.org/traceability/v1"</code>.</p>
-
-          <p>The object MUST have an <code>id</code> property, and its value MUST be a valid IRI (URI, URN).</p>
+          <p>It is RECOMMENDED the <code>id</code> be a valid UUID v4 URN, for example <code>urn:uuid:3978344f-8596-4c3a-a978-8fcaba3903c5</code></p>
```
OR13, line 1135: for cases like GS1, we cannot enforce a specific ID value, so I downgrade this to a recommendation

docs/index.html
```diff
@@ -1131,22 +1131,21 @@ <h3>Extensions to Standard</h3>
...
           <p>The object MUST conform to the basic requirements section of the <a href="https://www.w3.org/TR/vc-data-model/#basic-concepts">W3C VC Data Model</a>.</p>
+        </section>
 
+        <section>
+          <h4>Traceable Presentation</h4>
+          <p>This object extends <code>VerifiablePresentation</code> to support workflows.</p>
           <p>The <code>@context</code> MUST contain <code>"https://w3id.org/traceability/v1"</code>.</p>
-
           <p>The object MUST have an <code>id</code> property, and its value MUST be a valid IRI (URI, URN).</p>
+          <p>It is RECOMMENDED the <code>id</code> be a valid UUID v4 URN, for example <code>urn:uuid:3978344f-8596-4c3a-a978-8fcaba3903c5</code></p>
```
OR13, line 1148: move guidance on presentation id to `TraceablePresentaton` which is a type we define, and require support for.

docs/index.html
```diff
@@ -1247,50 +1246,115 @@ <h4>Workflow Instance</h4>
...
+          "postalCode":"65587",
+          "addressCountry":"Cyprus"
+        }
+    }
+  }
+}
+        </pre>
+        </section>
+        <section>
+          <h4>Workflow Replace</h4>
```
OR13, line 1294: This is the main new section, it supports a full update operation for a previous presentation.

docs/index.html
```diff
@@ -1247,50 +1246,115 @@ <h4>Workflow Instance</h4>
...
+          <p>
+A holder MAY use the <code>replace</code> property of a <code>TraceablePresentation</code>
```
OR13, line 1296: replace is an option member, that when present must be understood by verifiers who process the graph as linked data.

docs/index.html
```diff
@@ -1247,50 +1246,115 @@ <h4>Workflow Instance</h4>
...
+        </pre>
+        </section>
+        <section>
+          <h4>Workflow Replace</h4>
+          <p>
+A holder MAY use the <code>replace</code> property of a <code>TraceablePresentation</code> 
+to communciate to a verifier that a previous presentation should be replaced with a new presentation.
+          </p>
+          <p>
+Holders SHOULD ensure the previous presentation was accepted before presenting a <code>replace</code> presentation.
```
OR13, line 1300: Not a requirement, to support out of order message delivery, debatable

docs/index.html
```diff
@@ -1247,50 +1246,115 @@ <h4>Workflow Instance</h4>
...
+          </p>
+          <p>
+Holders MUST present the same credential types that were present in the previous presentation and no new credentials.
```
OR13, line 1303: Its a "replace" operation, not a partial update... this makes reasoning easier... another operation could be defined in the future for partial updates, but I don't recommend discussing this.

docs/index.html
```diff
@@ -1247,50 +1246,115 @@ <h4>Workflow Instance</h4>
...
+          </p>
+          <p>
+The presentation <code>id</code> of the <code>replace</code> presentation MUST be unique,
```
OR13, line 1306: presentation ids are always unique, even when they are for replacing a previous presentation by its id.

docs/index.html
```diff
@@ -1247,50 +1246,115 @@ <h4>Workflow Instance</h4>
...
+          </p>
+          <p>
+Holders MUST present the same credential types that were present in the previous presentation and no new credentials.
+          </p>
+          <p>
+The presentation <code>id</code> of the <code>replace</code> presentation MUST be unique, 
+and MUST NOT be the same as any previous values presented.
+          </p>
+          <p>
+The credential <code>id</code> of any corrected credentials MUST be unique,
```
OR13, line 1310: new credentials MUST be unique, this is debatable, and perhaps a problem for GS1

docs/index.html
```diff
@@ -1247,50 +1246,115 @@ <h4>Workflow Instance</h4>
...
+          <p>
+Holders MUST present the same credential types that were present in the previous presentation and no new credentials.
+          </p>
+          <p>
+The presentation <code>id</code> of the <code>replace</code> presentation MUST be unique, 
+and MUST NOT be the same as any previous values presented.
+          </p>
+          <p>
+The credential <code>id</code> of any corrected credentials MUST be unique, 
+and MUST NOT be the same as any previous values presented.
```
OR13, line 1311: This sentence is to protect the RDF reasoning over node identifiers.

docs/index.html
```diff
@@ -1247,50 +1246,115 @@ <h4>Workflow Instance</h4>
...
+    "https://w3id.org/traceability/v1"
+  ],
+  "id": "urn:uuid:p2",
+  "type": [
+    "VerifiablePresentation",
+    "TraceablePresentation"
+  ],
+  "replace": {
+    "id": "urn:uuid:p1",
+    "note": "I made a mistake in my presentation 'urn:uuid:p1', the invoice number was incorrect, this presentation corrects the error."
```
OR13, line 1330: note here, allows for the holder to explain themselves... this could easily create security or privacy issues, but it might also help with usability.

docs/index.html
mkhraisha: I think this should be worded in a way to include the case where only 1 credential needs to be replaced so only one credential is sent. 

The way it reads now is I must send every credential that was sent in the original presentation whether it needs replacement or not.

I think a better flow is for the replace operation to define a mapping that says replace credential id X with credential id Y, rather than replace presentation X with Y, because it enables the issuer to not have to keep track of previously sent presentations and keep the reasoning focused only on the credentials.

mprorock:
I would prefer something very explicit, like an instruction set for update, removal, other core operations.  those instructions can be in a presentation, but should target the credential level, not the entire presentation

docs/index.html
mprorock, line 1296:
```suggestion
A holder MAY use the <code>instructions</code> property of a <code>TraceablePresentation</code> 
```

docs/index.html
mprorock, line 1300:
```suggestion
to communciate to a verifier that previous credentials that were presented should be removed or replaced.
          </p>
          <p>
Holders SHOULD ensure the previous presentation was accepted before presenting an <code>instructions</code> presentation.
<p>Within `instructions` properties are included that are each an array of credential IDs
for which the property indicates the operation that should be performed on those credentials.
Defined in this specification are operations named `update`, and `delete`.
</p>
```

mprorock:
I am fine defining the vocab here - really not a fan of having business logic described normatively here - that seems like it should be described in the interop

brownoxford:
I am a big fan of @OR13 's suggestion to use JSON Patch here if possible.

TallTed:
A broad suggestion would be to change `instructions` to `instruction` (i.e., make it singular).

docs/index.html
OR13: ^ lets put this proposal in a separate PR, I prefer to address changes to the proposal in a follow on PR, its fine to suggest an issue marker.

OR13:
I think it would be better to do JSON PATCH as a second PR, so we have something to compare against.

nissimsan (approved):

BenjaminMoe (approved):

brownoxford (approved):

PR #818, opened 2023-06-20
Change the order of usmca from feedback

BenjaminMoe:
Updating the USMCA credential from feedback.

- Added invoice number in cases for a single instance instead of a blanket period
- Changed the order of the credential to closer resemble the paper-version of the document for business users
  - invoice number
  - exporter
  - producer
  - importer
  - product(s)
- removed producer confidential flag
- added various importers flag
- made exporter and producer direct objects instead of optional components

### DIFF
--- a/docs/openapi/components/schemas/credentials/USMCACertificationOfOrigin.yml
+++ b/docs/openapi/components/schemas/credentials/USMCACertificationOfOrigin.yml
@@ -91,6 +91,329 @@ properties:
           type: string
           enum:
             - USMCAProductSpecifier
+      invoiceNumber:
+        title: Invoice Number
+        description: >-
+          An invoice in the case of a single shipment. When covering a shipment, the
+          issued USMCA credential should be indicated as revokable as to not allow the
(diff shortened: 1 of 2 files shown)


OR13 (approved):

docs/openapi/components/schemas/credentials/USMCACertificationOfOrigin.yml
```diff
@@ -131,10 +456,17 @@ properties:
               title: Description
               description: Description of the shipment.
               type: string
+            gtin:
```
OR13, line 459: hopefully this produces JSON-LD that recognizes GS1 controls this term

BenjaminMoe: It's defined as `schema.org/gtin` in https://github.com/w3c-ccg/traceability-vocab/blob/main/docs/openapi/components/schemas/common/Product.yml#L26

nissimsan (approved):

PR #824, opened 2023-06-22
Add support for JSON Patch

OR13:
Spec text (would need a lot more to safely explain this).

Screen Shot 2023-06-22 at 10 06 50 AM

What a graph that contains operations looks like:

Screen Shot 2023-06-22 at 9 54 30 AM

What the simple solution we already have looks like

Screen Shot 2023-06-21 at 2 08 07 PM

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -1355,6 +1355,11 @@ had made a mistake in a previous presentation, and corrects the mistake with a r
   ]
 }
           </pre>
+
+          <section
+          data-include="sections/json-patch.html"
+          data-include-replace="true"
+        ></section>
         </section>
       </section>
     </section>
--- a/docs/sections/json-patch.html
+++ b/docs/sections/json-patch.html
@@ -1,0 +1,112 @@
+
+<section>
+  <h2>JSON Patch</h2>
+  <p>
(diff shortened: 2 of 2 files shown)


OR13:
I don't think we should proceed with this approach, it adds a lot of complexity to both holders and verifiers, and it destroys the value of graph queries over presented data.

mprorock:
> I don't think we should proceed with this approach, it adds a lot of complexity to both holders and verifiers, and it destroys the value of graph queries over presented data.

@brownoxford I think this is going to be way more complex than needed 

@OR13 perhaps there is a solution where we use the operations as defined by json patch, but provide the new credentials as payload?

mkhraisha:
I believe this PR is too complex, and the existing solution is fine until we can get consensus on a mandatory credential ID

OR13:
I am going to close this PR, the discussion has been had, and we should not take this path.

I will leave it open for 1 more pass before closing it.

nissimsan:
Not a viable direction. Direction on the call to abandon this. Let's take it up next week before dismissing.

OR13:
As discussed, we are closing this PR.

PR #825, opened 2023-06-23
Purchase order inlining

nissimsan:


docs/openapi/components/schemas/credentials/PurchaseOrderCredential.yml
```diff
@@ -49,7 +49,101 @@ properties:
   expirationDate:
     type: string
   issuer:
-    $ref: ../snippets/IssuerOrganization.yml
+    title: Issuer Organization
```
mkhraisha, line 52: We really should just have a common snippet for issuer org and use that. Instead of the way we've been in lining it

mkhraisha (approved):
Approve with nitpick about improvement

docs/openapi/components/schemas/credentials/PurchaseOrderCredential.yml
nissimsan: Cheers, @mkhraisha!

We're consciously moving away from `$ref: IssuerOrganization`: 

1. References in general makes schemas harder to consume and should be avoided. 
2. The kind of issuer organization isn't always the same and benefits from targeted descriptions and fields. Buyer issues PO, Seller issues CI, Shipper issues SI, Carrier issues BL, etc...

BenjaminMoe (approved):

OR13 (approved):

PR #833, opened 2023-07-05
Add product to intent to import

BenjaminMoe:
Updating product to reflect requirements for workflows. 
The expected behavior is that workflows are a series of credentials exchanged in order to complete a specific business flow. 

With respect to imports, the import needs to be scoped to importing one or more products contained within a shipment.


mprorock (approved):
I am good with this approach to resolve #833

mkhraisha (approved):
This makes sense to me, as is there is no way to associate a single Intent to Import with a workflow, that prevents replaying the same credential across many workflows.
This design means that there may be multiple intent to import credentials, one for each product in a workflow.

An alternative design might be to simply add a reference to the workflow ID within the intent to import and by doing so associate it to all products within the workflow.

nissimsan (changes requested):
Product, Commodity, and other should be presented along with - not within - the ITI credential. What is suggested here is not right direction to go: 
- Any change of existing business process is an obstacle to industry adoption.
- Overloading the ITI credential won't scale. 

Rather, let's introduce smaller snippet-size credentials such as ProductCredential and CommodityCredential which can be presented along with the ITI, bound by workflow instance.

PR #834, opened 2023-07-07
v2 data model prep

OR13:
Remove the v1 proofs, prepare to address context interop issues.


packages/traceability-schemas/services/contexts.js
```diff
@@ -1,204 +1,13 @@
 const fs = require('fs');
 const path = require('path');
-const cred = require('@transmute/credentials-context');
```
OR13, line 3: I bundled all the contexts we rely on in the contexts directory, and removed this vendor dep.

mkhraisha (approved):

nissimsan (approved):

PR #837, opened 2023-07-12
fix: req credentials bug

nissimsan:


nissimsan:
This fixes some of my own banal bugs. This affects no one else, so taking the liberty to merge out of band.

PR #842, opened 2023-07-25
Consolidate Transport and ShippingStop

rhofvendahl:
This PR consolidates `ShippingStop` and `Transport`, and changes examples and references as needed.

Why this seems appropriate:
- The two schemas have a large amount of overlap - the only additional field necessary to merge `ShippingStop` into `Transport` was `path`.
- I found `ShippingStop` confusing, as while it nominally refers to a location it has fields describing a leg of a journey, and is sometimes used that way.

The only tricky aspect of this is reconciling `Transport` as a path from a to b, and `ShippingStop` as a stop with additional information about the next leg of the journey. I went mostly with "it's a path", with some additional text in the Transport noting that in some contexts `departureLocation` may be referred to as a shipping stop.


TallTed (changes requested):
small stuff

rhofvendahl:
> This and #843 seem to have very similar changes to the workflows section

Both changes were part of build, I think. I assumed that was the intended behavior, though if it's specific to my setup I can look into it. The only change is removing double spaces so probably ok, right?

rhofvendahl:
@mkhraisha 

Looks like the big change in `workflows.html` is to remove `SVIP Steel Tech Demo` and replace it with `CBP Steel Tech Demo 2023`. Workflow schemas were changed to reflect this 2 wks ago with a merge commit by Orie (unsure of commit ref etiquette but `c3e51db`).

It looks as though for some reason `workflows.html` wasn't updated at that time. I'm not sure why it's updating for me, but it happens every time I run` npm run build` in the` traceability-vocab/packages/traceability-schemas` dir.

nissimsan:
Seems like the the html generation just have happened on your side. The content looks good.

nissimsan (approved):

OR13 (approved):

PR #843, opened 2023-07-25
Misc. schema property cleanup

rhofvendahl:
This PR cleans up a few small issues I've come across:
- `Product.productPrice` has a vague definition pulled directly from schema.org, whereas I'm pretty sure it should be the unit price
- `TransportEquipment.isShipperOwned` should be a boolean, not a string
- `MultiModalBoL` has a handful of copy/pasted terms that needed fixing
- `TransportEquipment` also has some issues with copy/pasted values

### DIFF
--- a/docs/sections/workflows.html
+++ b/docs/sections/workflows.html
@@ -34,6 +34,106 @@ mermaid: >-
       A->>B: present Verifiable Business Card</pre>
     
 
+      <h3>CBP Steel Tech Demo 2023</h3>
+      <pre class='mermaid'>sequenceDiagram
+  title CBP Steel Tech Demo 2023 Use Case: Steel import into USA from Mexico via road carrier
+  participant A as Steel Buyer
+  participant B as Steel Producer
+  participant C as Customs Broker
+  participant D as SIMA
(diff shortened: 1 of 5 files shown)


docs/sections/workflows.html
```diff
@@ -178,6 +134,41 @@ <h3>SVIP Steel Tech Demo</h3>
     C->>F: present Multimodal Bill Of Lading Credential</pre>
     
 
+      <h3>Intention to Import Workflow</h3>
+      <pre class='mermaid'>sequenceDiagram
+    participant A as Importer of Record
+    participant B as US Customs and Border Protection
+    A-->>A: issue Intent to Import Credential
+    A->>B: present Intent to Import Credential</pre>
+      <p>This simple workflow definition is used to file an Intent to Import  Credential in a Traceable Presentation and thereby start a US CBP Import  workflow. </p>
```
TallTed, line 143:
```suggestion
      <p>This simple workflow definition is used to file an Intent to Import Credential in a Traceable Presentation, and thereby start a US CBP Import workflow.</p>
```

OR13 (approved):

PR #854, opened 2023-08-10
Steel Tech Demo credential requirements

nissimsan:
SIMA Application is a means to the actual license, but should not be part of the workflow def requirements


BenjaminMoe:
I think SIMA license should have it's own workflow definition. And we should include a reference from this workflow to say, "this is where the SIMA license comes from"

nissimsan:
Agreed @BenjaminMoe, we have that: https://w3c-ccg.github.io/traceability-vocab/#us-sima-steel-import-license

mkhraisha (approved):

OR13 (approved):

PR #891, opened 2023-10-17
Set max locations for CTPAT locations

BenjaminMoe:
CTPAT locations should not be allowed to have an indefinite amount of locations. Capping at 100, which should be more than enough.


mkhraisha:
Where is this requirement coming from?

brownoxford (approved):

mkhraisha (approved):

OR13 (approved):

PR #910, opened 2023-11-27
Digital Product Passport links

nissimsan:
This adds ability to reference upstream credentials, enabling construction of a full product data graph


TallTed:
It would be good to title this without the acronym (or with both the acronym and its expansion), so `Digital Product Passport links` or `Digital Product Passport (DPP) links`.

mkhraisha (approved):

mkhraisha:
Can we create an issue to add the linked data terms?

nissimsan:
Issue raised, merging now as agreed.

PR #913, opened 2023-11-30
Measurement schema titles

nissimsan:
- Adds specific json schema titles to measurements
- Orders value first, unit after

### DIFF
--- a/docs/openapi/components/schemas/credentials/CommercialInvoiceCredential.yml
+++ b/docs/openapi/components/schemas/credentials/CommercialInvoiceCredential.yml
@@ -668,11 +668,11 @@ properties:
                         enum:
                           - QuantitativeValue
                     value:
-                      title: Value
+                      title: Weight Value
                       description: Value or amount.
                       type: string
                     unitCode:
-                      title: Unit Code
+                      title: Weight Unit
                       description: Unit of measurement.
                       type: string
                   additionalProperties: false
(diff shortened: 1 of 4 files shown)


docs/openapi/components/schemas/credentials/CommercialInvoiceCredential.yml
```diff
@@ -668,11 +668,11 @@ properties:
                         enum:
                           - QuantitativeValue
                     value:
-                      title: Value
+                      title: Weight Value
```
mkhraisha, line 671: Why Weight Value/ Weight Unit? I thought this was supposed to be generic to allow invoices of many types.

mkhraisha (changes requested):

docs/openapi/components/schemas/credentials/CommercialInvoiceCredential.yml
nissimsan: We want to be able to render documents without having to include the full hierarchy. So this is to distinguish for example Gross Weight from Net Weight in a flattened rendering. 

Note this is only the schema title - the key is still the generic `value`, so it's still wired into the generic class and its term definitions.

mkhraisha (approved):

OR13 (approved):

brownoxford (approved):

PR #923, opened 2024-01-10
Add `term` and `@id` for `serviceEndpoint`

PatStLouis:
Addresses [this issue](https://github.com/w3c-ccg/traceability-interop/issues/613).

Let me know if this reflects what we discussed.

Additionally, do we want to also include the following endpoints in the description?
```
      - https://platform.example/organization/123/credentials/status
      - https://platform.example/organization/123/presentations
```

I can bundle this into this pr.

@nissimsan @mkhraisha


mkhraisha (approved):

brownoxford:
Does this fix the issue you are seeing when you test locally with the new `traceability-v1.jsonld`? I notice that the document still contains the empty context for `TraceabilityAPI`:
 ```
        "TraceabilityAPI": {
            "@id": "https://w3id.org/traceability#TraceabilityAPI",
            "@context": {}
        },
```
and also that `serviceEndpoint` is already defined in [https://www.w3.org/ns/did/v1](https://www.w3.org/ns/did/v1).

PatStLouis:
@brownoxford @mkhraisha please be aware of [this pr](https://github.com/digitalbazaar/pyld/pull/189) on the `pyld` repository and feel free to chime in, we found a bug in the library which was evaluating empty contexts as `null` values, as is indicated by [section 5.1](https://w3c.github.io/json-ld-api/#overview-3) of the json-ld expansion algorithm. An additional review would be welcome!

mkhraisha:
Closed as resolved in a different library

PR #629, opened 2022-11-21
Ecommerce AEO program

nissimsan:
This PR updates schemas for an AEO program specifically for the new supply chain actors:
- Online Sellers
- Marketplaces
- Fulfillment Centers

At high level, there are three types of credentials in play: 
1. Organization Credentials: a) Marketplace know-your-seller, b) Seller AEO, c) Marketplace AEO, d) Fulfillment AEO
2. Product Registration Credentials: a) Seller Product listing, b) Fulfillment know-your-inventory, c) IPR 
3. Transaction Registration Credentials a) Order Registration, b) Fulfillment confirmation

Together, these enable linking shipments/transactions back to pre-registered product documentation submitted by program participants; with import pre-approval and AEO certificates as the main incentive mechanisms. 

The PR replaces some previous done Ecommerce schemas aiming for the same goals, but now aligns to our since established modelling conventions.

Issue #564, opened 2022-09-16
Product defects: ids (how to represent an individual product), quantities


VladimirAlexiev:
@nissimsan @mkhraisha cc @mgh128

I've looked at https://github.com/w3c-ccg/traceability-vocab/blob/main/docs/openapi/components/schemas/common/Product.yml. 
I'm comparing to EPCIS object events, which can use these standard IDs:
- `GTIN` for a product model
- `LGTIN` for GTIN plus a lot/batch number
- `SGTIN` for a GTIN plus a serial number

The identifiers in Product.yml are less well defined:
- `productID` gives example "ISBN", which suggests this should be used for a global identifier. But how would I know in which codification system?
- `sku` is a local part number. It's ok to have it, but doesn't provide global identification
- `batchNumber` corresponds to the `L` part of LGTIN. But there's no instruction whether it should be considered a suffix of `productID`
- `commodity` refers to a more generic code from the Harmonized System for customs purposes 
- there's no `serialNumber` so I cannot track individual expensive product like cars or airplanes

In addition, EPCIS events use a generic Quantity element that uses UNCEFACT Rec20 3-char UoM codes.
Product.yml has such generic property `schema:size` "sizeOrAmount", but also the 4 more specific `weight, depth, width, height`:.
- If I want to say "3kg of tomatoes" should I use `size` or `weight`?
- If I want to say "3 tomatoes", maybe `schema:size` would be a bit confusing?
- If I want to say "3 extra large tomatoes", should I go like this (turtle)
`<product> schema:size "extra large", 3`
or maybe even
`<product> schema:size "extra large", [a schema:QuantitativeValue; value 3]`

Maybe these specific props just cause confusion?

mkhraisha:
@VladimirAlexiev, I don't think we have thanked you enough for your wonderful work. Thank you for this, and the many other issues you opened.

`productID` is intentionally made to be a generic id, that may or may not be a global identifier, some of the products in this vocabulary (such as a barrel of oil), simply do not have a global id in the industry, rather, they are provided IDs by individual companies that created them.

`batchNumber` Excellent point, if we choose to make productID more specific we can align it there. again looking at the barrel of oil example, `batchNumber` is used as an identification to identify batches of barrels of oil ( usually in the 60k range) and the number is issued by the pipelines, each pipeline creates its own numbering system. Pipelines also regularly reuse batch numbers and combine them with a date of creation to ensure uniqueness.

[...]

PR #927, opened 2024-01-23
Remove one reference sizeOrAmount

nissimsan:
This removes one reference to `sizeOrAmount` as pointed out on https://github.com/w3c-ccg/traceability-vocab/pull/629. 

However, I can't remove the actual definition on Product.yml because it is heavily references by Agri and Oil examples - beyond what I feel I should be PR'ing. I do encourage getting rid of these, though, so we can remove the poorly defined `sizeOrAmount` from Product altogether.


TallTed:
s | https://github.com/w3c-ccg/traceability-vocab/pull/629 | https://github.com/w3c-ccg/traceability-vocab/pull/564 |

PR #940, opened 2024-03-08
Editor list update

nissimsan:
@OR13 asked that I replace him as editor.

### DIFF
--- a/docs/credentials-with-issuer-dependent-terms.json
+++ b/docs/credentials-with-issuer-dependent-terms.json
@@ -68,7 +68,7 @@
     "count": 0
   },
   {
-    "type": "OrganicCertificationCredential",
+    "type": "OrganicCertificateCredential",
     "count": 0
   },
   {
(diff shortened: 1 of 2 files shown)


mkhraisha (approved):

docs/credentials-with-issuer-dependent-terms.json
```diff
@@ -68,7 +68,7 @@
     "count": 0
   },
   {
-    "type": "OrganicCertificationCredential",
+    "type": "OrganicCertificateCredential",
```
TallTed, line 71: This seems out of place for this PR ... though it is certainly a change that should be made.

nissimsan: Definitely - my bad

rhofvendahl: This is due to a recent change, but does need to be merged in. Seems like it'd be fine to include this and merge.

nissimsan: OK, thanks @rhofvendahl !

Issue #593, opened 2022-10-14
Date string should be switched to Datetime


BenjaminMoe:
https://github.com/w3c-ccg/traceability-vocab/blob/main/docs/openapi/components/schemas/common/CTPAT.yml#L58

```yml
  dateOfLastValidation:
    title: Date of Last Validation
    description: >-
      Date of last validation.
    type: string
    $linkedData:
      term: dateOfLastValidation
      '@id': https://schema.org/Date
```

Value is a Datetime, should use Datetime for JSOn Schema 
`"dateOfLastValidation": "2022-01-06T11:50:00Z",`

TallTed:
Please confirm whether this is really just a DATE (as the name suggests) or really a DATETIME (as the sample value suggests) when recorded on existing paper forms and/or electronic systems.

If the latter, and if possible, I would suggest changing the attribute from `dateOfLastValidation` to `timeOfLastValidation` or perhaps better `datetimeOfLastValidation` or `timestampOfLastValidation`.  Even if `dateOfLastValidation` cannot be changed, if the value really is meant to be a datetime/timestamp, I would suggest changing the title, description, and any other human-focused info to reflect this -- i.e., --
```pre
title: Datetime of Last Validation
description: >-
   Datetime of last validation.
```
-- or --
```pre
title: Timestamp of Last Validation
description: >-
  Timestamp of last validation.
```

brownoxford:
Discussed on call, confirm that field is datetime not just date. We would want to be more specific and define this as a datetime and not just a string. We also want to update the title to indicate that time is involved, not just date.

mkhraisha:
Action item: Update top level credentials to use DateTime where relevant. Do not update individual stream credentails

mgh128:
GS1 EPCIS 2.0 casts values of its `eventTime` and `recordTime` fields to an xsd:dateTimeStamp data type (see https://www.w3.org/TR/xmlschema11-2/#dateTimeStamp ), which requires an explicit time zone offset to be specified, to avoid ambiguity on a global basis about when an event occurred, especially if shipments might cross multiple time zones.

mkhraisha:
@rhofvendahl Please review the agriculture credentials for any date issues.
@mkhraisha please review the OG credentials for any date issues

rhofvendahl:
@mkhraisha did we resolve to default to date & only use datetime if time is needed? (meeting notes weren't uploaded for that one and my own notes are spotty)

[...]

PR #942, opened 2024-03-11
Clean up date and datetime usage

rhofvendahl:
This PR updates term IDs which refer to date/datetime/time as discussed in https://github.com/w3c-ccg/traceability-vocab/issues/593, and also adds appropriate XSD datatypes for additional clarity (as in many cases otherwise relevant term IDs aren't good indicators of the actual type expected) as suggested in https://github.com/w3c-ccg/traceability-vocab/issues/213.

NOTE: in a lot of cases I used https://schema.org/endDate as the term ID. There isn't a property https://schema.org/date, and I think it's fine to have the implication that, in any case where an activity takes some significant duration of time (like harvesting or inspection), the relevant date/datetime should be when the activity is finished. In a few cases where it's clear the property refers to a start of an activity I used https://schema.org/startDate.

Further discussion of XSD datatypes:
https://github.com/schemaorg/schemaorg/issues/1781
https://github.com/schemaorg/schemaorg/issues/1748

### DIFF
--- a/docs/openapi/components/schemas/common/AgricultureActivity.yml
+++ b/docs/openapi/components/schemas/common/AgricultureActivity.yml
@@ -47,7 +47,8 @@ properties:
     type: string
     $linkedData:
       term: activityDate
-      '@id': https://schema.org/DateTime
+      '@id': https://schema.org/endDate
+      '@type': http://www.w3.org/2001/XMLSchema#date     
   activityType:
     title: Activity Type
     description: Type of Activity, e.g., plant, harvest, spray.
--- a/docs/openapi/components/schemas/common/AgricultureInspectionCommonInfo.yml
+++ b/docs/openapi/components/schemas/common/AgricultureInspectionCommonInfo.yml
@@ -61,6 +61,7 @@ properties:
     $linkedData:
       term: inspectionStarted
       '@id': https://schema.org/startDate
+      '@type': http://www.w3.org/2001/XMLSchema#dateTime
   inspectionEnded:
     title: Inspection Ended
     description: Date and time inspection ended in ISO 8601 format (e.g. 1970-01-01T00:00Z for UTC, or 1970:01:01T00:00-5:00 for New York on standard time).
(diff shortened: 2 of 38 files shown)


docs/openapi/components/schemas/common/AgricultureActivity.yml
```diff
@@ -47,7 +47,8 @@ properties:
     type: string
     $linkedData:
       term: activityDate
-      '@id': https://schema.org/DateTime
+      '@id': https://schema.org/endDate
+      '@type': http://www.w3.org/2001/XMLSchema#date
```
OR13, line 51: this RDF type is not correct i think see

https://github.com/w3c/vc-data-model/blob/main/contexts/credentials/v2#L119

OR13 (changes requested):
incorrect RDF type

docs/openapi/components/schemas/common/AgricultureActivity.yml
rhofvendahl: @OR13 do you mean that this is `http://www.w3.org/2001/XMLSchema#date` & not `http://www.w3.org/2001/XMLSchema#dateTime
http://www.w3.org/2001/XMLSchema#date`? If so that's intentional, I think this one's not supposed to have a time attached.

Unless I'm misremembering the approach we settled on for the 2/6 trace meeting? Did we say that we want all instances of "date" to be "datetime" but with the implicit assumption that if there's no specified time it's midnight? We neglected to upload notes that meeting & my own are kinda spotty 😅

nissimsan: #date is the correct granularity.

nissimsan: And endDate is better than DateTime... Not good, but better

TallTed: `validUntil` is appropriately a datetime.

`activityDate` is appropriately a date.

nissimsan (approved):

mkhraisha (approved):

Issue #596, opened 2022-10-18
Do we make `id` a requirement for trace-vocab credentials?


BenjaminMoe:
Issue spawned from PR: https://github.com/w3c-ccg/traceability-vocab/pull/592

`id` is optional in the VC-data model. For trace-vocab, we're expecting the RDF graph representation of our credentials to be interoperable. We want to force good JSON-LD practices and not have empty nodes.

Proposal:
1. We add `id` as a required value to all of the credentials defined in trace-vocab
2. We make the requirement and difference clear in the respec document.

BenjaminMoe:
By definition of profile, we expect `id` to exist in all issued credentials.

OR13:
We expect `id` to be required in all schemas for credentials defined by Trace Vocab... we cannot assert that for foreign credentials / GS1 Credentials, etc...

mkhraisha:
We should do this across the board:

Assigning @nissimsan  to tackle the steel credentials.
Assigning @mkhraisha  to tackle the oil credentials.
Assigning @rhofvendahl to tackle the Ag credentials.
Assigning @paulfdietrich  to tackle all the GS1 credentials.

paulfdietrich:
An ID is required in our GS1 IdentificationKeyLicense, Prefix, Company Prefix, and Key credentials. The definition of a GS1 Data credential would allow the omission of a credential ID, but I believe we could simplify and require this ID for all of the GS1 credentials within this vocab.  I opened a ticket in the GS1 data model repo to clarify the language around the requirements for the credential ID>

mkhraisha:
This should be as simple as adding required to the yaml file for the credentials for the ID attribute

mkhraisha:
@mkhraisha @nissimsan @rhofvendahl @paulfdietrich we should get this in ASAP

PR #960, opened 2024-06-27
Mandatory ids on steel and ecommerce

nissimsan:
Makes some missing mandatory id where missing on ecomm and steel schemas.

https://github.com/w3c-ccg/traceability-vocab/issues/596


nissimsan:
Oh, @mkhraisha you did them all on https://github.com/w3c-ccg/traceability-vocab/pull/958. Closing this then

Issue #887, opened 2023-10-10
"TraceAPI" is not a workflow actor


nissimsan:
Hey @rhofvendahl, I think [this](https://github.com/w3c-ccg/traceability-vocab/blob/main/docs/openapi/components/schemas/workflows/common-import-credentials.yml#L25) is a mistake. TraceAPI is not a supply chain actor.

rhofvendahl:
@nissimsan This example is intended to illustrate credentials being passed from various parties to mesur.io's VC-API and then from there to CBP (with some addnl steps). Does that sound like a reasonable usage?

nissimsan:
@rhofvendahl , in for example `A->>C`, I would argue that `->>` _is_ TraceAPI.

nissimsan:
@rhofvendahl, as discussed on the call, I suggest deleting the workflow altogether. It seems in conflict with the "US CBP Entry" workflow which is the always mandatory documents required for entry filings.

PR #962, opened 2024-08-08
Remove common import creds workflow

rhofvendahl:
This PR removes `workflows/common-import-credentials.yml`, as discussed in https://github.com/w3c-ccg/traceability-vocab/issues/887


mkhraisha (approved):
This is fine overall, Nis will be rebasing this PR to fix the extraneous addition of the apparel HTML

docs/sections/workflows.html
```diff
@@ -34,6 +34,250 @@ <h3>Business Card Workflow</h3>
       A->>B: present Verifiable Business Card</pre>
     
 
+      <h3>CBP E-Commerce Apparel Tech Demo 2024</h3>
```
nissimsan, line 37: This should not be part of this PR.

nissimsan (approved):

brentzundel (approved):

docs/sections/workflows.html
nissimsan: Understood, this was never build into main (only gh pages). Fine then.

PR #966, opened 2024-08-15
modify editors

mkhraisha:
Updates Brent/Nis Editors


nissimsan:
I think you mean Brent/Mike editors, @mkhraisha :)

PR #977, opened 2024-09-24
Add data about packaging

brentzundel:
this PR adds a common schema for Instrument of Trade, which is a term of art for packaging material.

### DIFF
--- a/docs/openapi/components/schemas/common/InstrumentOfTrade.yml
+++ b/docs/openapi/components/schemas/common/InstrumentOfTrade.yml
@@ -29,7 +29,6 @@ properties:
     description: >-
       IPPC ISPM 15 stamp information as described in
       https://www.ippc.int/en/publications/640/
-    type: object
     properties:
       originCountry:
         title: Country
--- a/docs/openapi/components/schemas/common/InstrumentOfTrade.yml
+++ b/docs/openapi/components/schemas/common/InstrumentOfTrade.yml
@@ -26,7 +26,8 @@ properties:
   ippcStamp:
     title: IPPC Stamp
     description: IPPC Stamp information
-    type:     description: >-
+    type: object
+    description: >-
       IPPC ISPM 15 stamp information as described in
       https://www.ippc.int/en/publications/640/
     type: object:
--- a/docs/openapi/components/schemas/common/InstrumentOfTrade.yml
+++ b/docs/openapi/components/schemas/common/InstrumentOfTrade.yml
@@ -1,0 +1,77 @@
+$linkedData:
+  term: InstrumentOfTrade
+  '@id': https://w3id.org/traceability#InstrumentOfTrade
+title: Intrument of International Trade
+description: >-
+  The actual packaging (container, crate, pallet, etc.) which is used to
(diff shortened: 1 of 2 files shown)


mkhraisha (approved):

docs/openapi/components/schemas/common/InstrumentOfTrade.yml
```diff
@@ -0,0 +1,76 @@
...
+          NPPO to the producer of the wood packaging material or treatment
+          provider who applies the marks or the entity otherwise responsible
+          to the NPPO for ensuring that appropriately treated wood is used and
+          properly marked (shown in the examples as “000”). The number and order
+          of digits and/or letters are assigned by the NPPO.
+        type: string
+      treatmentCode:
+        title: Treatment Code
+        description: >-
+          a treatment code using the appropriate abbreviation according to ISPM
```
TallTed, line 50:
```suggestion
          A treatment code using the appropriate abbreviation according to ISPM
```

TallTed (approved):

nissimsan:
@brentzundel , "InstrumentOfTrade" seems like a re-invention of https://vocabulary.uncefact.org/LogisticsTransportEquipment . We already have this on Trace: 
https://w3c-ccg.github.io/traceability-vocab/#TransportEquipment . I recommend using it instead, extending as needed.

brentzundel:
@nissimsan Transport Equipment seemed to be shipping containers and trucks, rather than pallets and boxes. Are they really the same thing? If so I can extend transport equipment to include packing material and wood treatment stamps, but they seemed at different layers to me.

nissimsan:
@brentzundel , 

You commonly distinguish returnable vs non-returnable. 

Transport Equipment is returnable. This includes pallets: 
https://vocabulary.uncefact.org/containedTransportEquipment
"A piece of transport equipment contained within this piece of logistics transport equipment, such as a pallet."

If you are looking for non-returnable, consider: 
https://vocabulary.uncefact.org/Package

It doesn't seem like it should be necessary to invent a term here.

mprorock (approved):

TallTed:
@nissimsan — It seems worth noting that pallets are often *not* returnable (or at least, not *returned*), leading to endless ads for "[FREE PALLETS](https://www.google.com/search?q=free+pallets)!"

It seems that a loaded-and-wrapped pallet in a "[shipping container](https://www.google.com/search?q=shipping+container)" that gets loaded from dock to ship to railcar to semi-trailer *could* be interpreted as *either* a [Package](https://vocabulary.uncefact.org/Package) _or_ a piece of [containedTransportEquipment](https://vocabulary.uncefact.org/containedTransportEquipment), by the linked definitions. 

(I find it odd that they have [contained Transport Equipment](https://vocabulary.uncefact.org/containedTransportEquipment), but no [***un***contained Transport Equipment](https://vocabulary.uncefact.org/uncontainedTransportEquipment).)

All of which to say... UNCEFACT's ontology seems to be less than fully thought through, or at least, less than clearly documented. Confusion seems to be built in.

nissimsan (approved):

